PR #79, opened 2022-12-19
Be stricter about the camera restarting after a property change.

ehpor:
The ZWO camera kept on crashing after exposure time changes. This is a bit slower, but as we're not changing exposure times that often, this is a worthwhile tradeoff to make.

The camera is not stopped for every property change, except for offset_x and offset_y.

Tested on hardware.


raphaelpclt:
Something got fixed in the meantime, not really happening anymore. Closing this issue.

PR #85, opened 2023-02-02
Move to Python 3.9.12.

ehpor:
Should be merged together with https://github.com/spacetelescope/hicat-package2/pull/170. Main discussion points to be put in that PR.


a-sevin:
Just a comment linked to conda not to the python version, do you know [mamba](https://github.com/mamba-org/mamba)? It's a reimplementation of the conda package manager in C++. It's very fast, I use it since last summer on several linux servers and it's really nice.

PR #124, opened 2023-10-17
Bugfix/sci cam sign flip

steigersg:
See affiliated PR on hicat-package2 (https://github.com/spacetelescope/hicat-package2/pull/480)

Add infrastructure in camera services to rotate and/or flip the images on the camera.

The camera rotation happens first and defines which camera axis (width or height) is aligned with the gravity vector of the testbed. Defining `rot90` to be `True` will perform one counter-clockwise 90 degree rotation. If a clockwise rotation is desired then `flip_y` should also be set to `True`.

The flips (`flip_x`, and `flip_y`), with the exception of the example above, are meant to be defined with respect to the optical axis (e.g. whether a lens upstream of the camera has flipped the image). 

Notably `offset_x` and `offset_y` values are not changed with the reflections and rotations since these are fundamentally defined with respect to the camera array and the setters and getters for these values are directly given to the setROI functions in the vendor camera control software itself. I believe this is alright since finding these values should not have to occur very frequently and the user can take care to make sure they are defined with respect to the coordinate system of the camera itself.

## Tasks before review

- Test on hardware (might have to do with phase retrieval camera since this is the only current one with a rotation)
- Optimize for speed and memory usage

### DIFF
--- a/catkit2/services/flir_camera/flir_camera.py
+++ b/catkit2/services/flir_camera/flir_camera.py
@@ -133,6 +133,9 @@ class FlirCamera(Service):
         self.height = self.config['height']
         self.offset_x = self.config['offset_x']
         self.offset_y = self.config['offset_y']
+        self.rot90 = self.config.get('rot90', False)
+        self.flip_x = self.config.get('flip_x', False)
+        self.flip_y = self.config.get('flip_y', False)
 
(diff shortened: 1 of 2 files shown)


raphaelpclt:
Should be superseded with camera base class #131

ivalaginja:
Superseded by #131.

PR #131, opened 2023-10-24
Add base class for cameras

ehpor:
WIP but I might want some feedback soon.

~These base services generalize much of the functionality that is currently duplicated between cameras and DMs. This simplifies their implementation and makes fixing bugs in ROI flips and rotations, DM channel updates, etc. much easier in the future.~

The DM base class has been moved out of this and merged a while ago.

This PR introduces a camera base class that generalizes much of the functionality that is currently duplicated between all the different camera services. This simplifies their implementation and makes finding bugs in ROI flips and rotations in particular much easier. This work is entirely based on what's in https://github.com/spacetelescope/catkit2/pull/124.

Note how successfully merging this PR should result in:
- Closing without merging of https://github.com/spacetelescope/catkit2/pull/124
- Reevaluation and eventual merging of https://github.com/spacetelescope/hicat-package2/pull/480
- Reverting https://github.com/spacetelescope/catkit2/pull/132


ivalaginja:
Just a note: the DM base classes have now been merged in #156, so anything DM in this PR can simply be removed.

ivalaginja:
I will be taking this over from @alexisyslau.

Considering the discussion in https://github.com/spacetelescope/catkit2/pull/124, the three parameters `rot90`, `flip_x` and `flip_y` are sufficient to transform images from camera coordinates to system coordinates. I am now trying to figure out how to choose these three parameters based on measured images on the THD2 testbed, so I can determine our system-level coordinate system for images.

Choosing rot90 should be easy because that is figuring out what is up-down, and what is left-right. For the flips, I understand they should be used to "flip back" the image that would have flipped by an upstream lens, but I wonder how to determine my initial reference? Is it just that if I have an even number equal lenses upstream, I don't need a flip, and if I have an uneven number of equal lenses upstream, I do need the respective flip? Thoughts @steigersg @ehpor?

Edit: Could I also assume that the goals of the flips in x and y are to simply match the optical model?

ivalaginja:
Spent a good chunk of time on this with @steigersg today and made good progress. What's left to do:
 - [x] Rewrite the offset transformation matrix to accommodate an inverse transform; also check the forward logic.
 - [x] Test the transformation function on arrays outside of the service.

ivalaginja:
@ehpor do you think you could look over this for an initial review? I would like to get some feedback on the base class and on the ZWO v2, as I want to have something robust here first before moving on to the other cameras. We also have none of the ZWO or FLIR over here, so before I can test anything, I will need to do the transfer also for the AVT and/or Hamamatsu cameras first.

ivalaginja:
@ehpor @steigersg @alexisyslau I'd like to ask your guys' advice because this PR is slowly breaking my brain. @steigersg and I figured out the different transformations and property logic (we think), which is the first big part of this PR. The second big part is to actually use the camera base class with one of the camera and check the beforementioned transformations and logic.

Since it looks like I am the one moving forward with this, I need a new class of a camera type that I have access to on THD2, which is now the Hamamatsu ORCA.

Could I ask you guys for feedback specifically on the new file `hamamatsu_camera_v2.py` and whether this looks the way it should? I am struggling in particular with the following points:
- For the Hamamatsu, we need to read the sensor width and height before changing the ROI (I forgot why we can't read this from the config but there was a reason. I can go dig for it if this becomes a pain). In the old Hamamatsu service this is simply assigned to a class attribute in the `open()`, but with the base class this would have to be done smarter and right now I don't know how.
- Inside the Hamamatsu service, we override the `monitor_temperature()` method because we added a safety check inside, I don't know if there's a way to duplicate things less than I am doing here since it uses a `while` statement.
- The hardest part for me right now is to figure out how to work in the `StoppedAcquisition` context manager correctly. Almost all Hamamatsu properties need that, with only 2-3 exceptions.

Any help is appreciated.

ivalaginja:
@alexisyslau and I solved the first two of the above points. The last one still gives me headaches.

The basic camera properties are defined in the camera base class. It is also in the camera base class that they are made into catkit2 properties (using the `make_property_helper` that calls `self.make_property`). The problem I am facing is that I have a bunch of these basic camera properties, i.e. width, height, offset x and offset y, that need to use the `StoppedAcquisition` context manager. This context manager is not used inside the camera base class though, so I don't know how to implement this in the Hamamatsu child class so that it applies to these camera properties.

It seems to be that by defining the basic properties like width and height etc. in the base class, I loose the option to warp them into a `StoppedAcquisition` optional.

@ehpor please help?

ivalaginja:
The above is now solved, @RemiSoummer can give more details on that.

I have been able to run the Hamamatsu v2 service on hardware: I can stream images on the viewer and I can change the exposure time successfully. However, changing the offsets gives me errors so I will stop here for now and try to come back to it with a fresh head another time.

ivalaginja:
@steigersg @RemiSoummer I have now written the ZWO v2 service in exactly the same way like the Hamamatsu service that I started testing on hardware. Depending on your availability, it is ready to be tested on hardware too. Note how I remove the HiCAT hack in the ZWO service that should be fixed through this PR and the correct config parameters for the camera.

RemiSoummer:
After discussion with @ivalaginja (sending this message after the fact), documenting here a few thoughts. 
The issue is that the Hamamatsu has some properties that require stopping the acquisition.  First this sounds to me like a reasonable feature that may exist on other types of cameras so I'm interested in solving this in a general way, even if this means for ZWO users we have to change something on our end.  

- My first question was whether we can override the `make_property_helper` at the child class level to add the logic and the context manager to turn off the acquisition for the Hamamatsu (or other brands in their own future child classes).  We agree this would be a bad solution because this defeats the purpose of base class. Indeed we need the acquisition pause for properties that require the flip/rotation logic, so we would be over-writing those functions that contain the rotation/flip logic, therefore having to duplicate that critical code.  The point of the base class is that we never touch the code that handles the rotation/flips between the detector frame of reference to the testbed frame of reference. 

- Another solution would be to implement some kind of decorator with the acquisition pause context manager and logic around the setter/getters that call the API in the child service.   I don't like this solution because 1) not sure how to implement that properly and 2) this means the code would look different depending on the camera and by looking at the ZWO service one does not know how to implement another brand that requires the acquisition pause without looking at the Hamamatsu example.  I prefer a solution where the code is the same for all cameras for basic functionalities and we use some kind of config entries to turn on or off standard features (I'm considering that the acquisition pause as something that is probably standard for a number of brands, as opposed to stuff very specific to the Hamamatsu e.g. the fan.)

- The chosen solution is to have class attributes that are hard coded in the service child class, for example something like `self.offset_x_requires_acquistion_pause = True` and same for all other properties.  The make_property_helper would then have an optional parameter e.g. `requires_acquistion_pause` that defaults to `False` and called by the child service.     An alternative would be to put additional parameters in the config file but I prefer moving those at the child service levels because they depend on the hardware, and that's something we will hard code in at the same time we have the camera manual and API in front of us.  Adding those parameters `offset_x_requires_acquistion_pause = True/False` at the config file level gives the chance the user messes with it and it's really strictly limited to the hardware so we don't need to expose that to the user.   With the chosen solution, the base class is untouched and the code is identical for all subclass implementations, we don't expose anything unnecessary at the config file, and it just requires setting up these flags for each properties in the child service.  Moreover, we can do that easily when we migrate the ZWO and FLIR services to use the base class, so this does not seem too invasive to me.

RemiSoummer:
> @steigersg @RemiSoummer I have now written the ZWO v2 service in exactly the same way like the Hamamatsu service that I started testing on hardware. Depending on your availability, it is ready to be tested on hardware too. Note how I remove the HiCAT hack in the ZWO service that should be fixed through this PR and the correct config parameters for the camera.

@ivalaginja I think we could also do this in two PRs maybe?  Since you have time pressure to merge the base class and the Hamamatsu service @steigersg and I can do the ZWO transition later in a separate PR and we still use the current service in the mean time.

steigersg:
> > @steigersg @RemiSoummer I have now written the ZWO v2 service in exactly the same way like the Hamamatsu service that I started testing on hardware. Depending on your availability, it is ready to be tested on hardware too. Note how I remove the HiCAT hack in the ZWO service that should be fixed through this PR and the correct config parameters for the camera.
> 
> @ivalaginja I think we could also do this in two PRs maybe? Since you have time pressure to merge the base class and the Hamamatsu service @steigersg and I can do the ZWO transition later in a separate PR and we still use the current service in the mean time.

@ivalaginja @RemiSoummer If this PR just requires a test on the ZWO to be fully done then I can prioritize doing that this week (likely Friday) just so it's off our plates. If we think it requires more detailed tests (or there's any remaining outstanding issues) though then I agree with Rémi that we should split it up so that it's not blocking Iva.

FWIW not having looked at in in super great detail I agree with implementing the last bullet point Rémi posted especially since we already have a concession in the `make_property_helper` for `requires stopped acquisition` in the ZWO.

ivalaginja:
Slowly advancing with hardware tests with the Hamamatsu v2. Very stuck now though after trying to run on hardware with @alexisyslau and getting the following error when I try to change one of the offsets:

Screenshot 2025-09-25 at 15 42 45

This seems to be a problem in the context manager that stops and restarts the acquisition, as the `end_acquisition()` never takes fully effect and then the camera call times out (the numerical -2.... is a timeout error). We confirmed that changing the offset x works fine with the original Hamamatsu service.

Any ideas what to do about this? @ehpor?

ivalaginja:
Alexis figured it out, the StoppedAcquisition context manager needs to call on the hidden methods `_end_acquisition()` and `_start_acquisition()` instead of the camera-specific ones (`end_acquisition()` and `start_acquisition()`).

ivalaginja:
So now it works - but only for offset changes in multiples of 100 😅 meaning 100, 200, 300, ... which is probably simply the next problem ^^

ivalaginja:
Bote how the current Hamamatsu v2 does not drop the very first frame in an acquisition like the original service does. This still needs to be implemented.

RemiSoummer:
I am struggling to try the v2 for the zwo_camera.   I changed the service_type to v2 in the config file, fresh compilation of catkit and fresh server.   

image

Screenshot 2025-09-25 at 1 27 42 PM

it's not finding the v2.  

One thing also that I do not understand its that in the catkit2 toml file we have 
`zwo_camera = "catkit2.services.zwo_camera.zwo_camera"`

I tried updating that to v2 and same error, plus @ivalaginja on THD does not have to put the v2 there for the hamamatsu_v2 to work, so I'm at a loss understanding what's happening here, any idea?

steigersg:
Made a bit of progress - I changed the zwo_camera line in the `pyproject.toml` entry points to be

`zwo_camera_v2 = "catkit2.services.zwo_camera_v2.zwo_camera_v2"`

and that allowed the service to be found (woohoo). I'm getting some other errors now (see attached) but these are related to the actual changes in the zwo_camera_v2 service itself so at least we're making forward progress. 

Changing the name and location of the service entry point does make sense to me as something that would need to be done so I'm still quite confused as to how Iva was able to run the Hamamatsu v2 without updating her entry point, but that may be due to some cache weirdness or something? (pure speculation). 

Screenshot 2025-09-25 at 3 53 09 PM

ivalaginja:
I blame cache weirdness but @alexisyslau reports being able to start the sim camera v2 without a toml change as well so I'm still confused.

alexisyslau:
> I blame cache weirdness but @alexisyslau reports being able to start the sim camera v2 without a toml change as well so I'm still confused.

@ivalaginja I could not recall how I do that in the past but I now changed the toml file by editing the entry point, and re-install catkit2. It did run in the simulator and I can access the service, and not crashing server.

alexisyslau:
> Made a bit of progress - I changed the zwo_camera line in the `pyproject.toml` entry points to be
> 
> `zwo_camera_v2 = "catkit2.services.zwo_camera_v2.zwo_camera_v2"`
> 
> and that allowed the service to be found (woohoo). I'm getting some other errors now (see attached) but these are related to the actual changes in the zwo_camera_v2 service itself so at least we're making forward progress.
> 
> Changing the name and location of the service entry point does make sense to me as something that would need to be done so I'm still quite confused as to how Iva was able to run the Hamamatsu v2 without updating her entry point, but that may be due to some cache weirdness or something? (pure speculation).
> Screenshot 2025-09-25 at 3 53 09 PM

What happened before tho? Were you trying to set attributes? But it looks like it just failed when you are trying to set the offset_x. Idk if your attributes are being set? I think your function is expecting something like int but it got something else.

ivalaginja:
> So now it works - but only for offset changes in multiples of 100 😅 meaning 100, 200, 300, ... which is probably simply the next problem ^^

I just found in our internal test report for the Hamamatsu that the offsets, like width and height, need to be numbers divisible by 4. I'll test this again on hardware and if this is confirmed, I can finally start testing the transformation logic of the flips and rotations 🤞

ivalaginja:
Hamamatsu v2 service fully works! I am able to set and read the various properties correctly too. My next goal is to test the rotation and flip logic. @steigersg I might come back to you about this 😅

catkit2/base_services/camera.py
```diff
@@ -0,0 +1,399 @@
...
+        # Perform the dot product.
+        # Translate to ROI center, rotate, flip in x then in y,
+        # and finally translate back to origin.
+        coords = [x, y, 1]
+        new_coords = np.linalg.multi_dot([T_back, Y, X, R, T_center, coords])
+
+        if inverse:
+            new_coords = np.linalg.multi_dot([T_back, Y, X, R.T, T_center, coords])
+
+        return new_coords[0], new_coords[1]
```
steigersg: Think this needs to return an `int` for each new coordinate. This returning a float was causing the TypeError I posted about in the conversation on this PR. Casting these two values to an int allows me to open a camera running the `zwo_camera_v2` service and start taking images.

steigersg:
With the type fix change I just pushed from the makidonlab account I can run the `zwo_camera_v2` from the HiCAT PC without crashes. As Iva said though still need to confirm that all the transformation logic is sound.

catkit2/base_services/camera.py
```diff
@@ -0,0 +1,399 @@
...
+
+        if self.flip_y:
+            # Define y reflection matrix.
+            Y[1][1] = -1
+
+        # Define translation matrix back so that the origin is in the upper left as expected.
+        T_back = np.eye(3, 3)
+
+        # If forward with rotation, or if inverse without rotation.
+        if (self.rot90 and not inverse) or (not self.rot90 and inverse):
```
steigersg: Looking at this again I don't think the second logic point is necessary... if we're not rotating then I don't know why the height and width would need to get swapped for the back translation

steigersg:
@ivalaginja made two changes to the `transform_offset` function: 
- replaced `height` and `width` with `sensor_height` and `sensor_width` since the offset is the location of the ROI with respect to the detector 
- In the inverse transform moved the `R.T` matrix to after the `X` and `Y`reflections. Confirmed for all iterations of `rot90`, `flip_x` and `flip_y` that running the transform and inverse transform now returns to the same coordinates.   

Still want to test these changes on hardware but this function is now behaving in all the ways I would expect at least numerically.

steigersg:
@ivalaginja I found some bugs in the offset transform in particular when the 90 degree rotation is enabled. A summary of what was added:

- Added hidden variables for offset_x and offset_y so that those values can be internally retrieved without calling the getter. This is important for the 90 degree rotation case which requires knowledge of both the x and y offsets even if the user is only requesting one. 
- In each of the offset_x and offset_y setters, the call to the set_roi_offset function is made for both x and y even if only one of those values is updated. This is again for the 90 degree rotation case since moving in one dimension in user coordinates will move the roi in the opposite dimension in camera coordinates necessitating both being updated . 

I tested this on the new HiCAT ZWO camera but would appreciate you also testing this out if possible/ double checking the logic (especially when rot90 = True). As you know it can be tough to keep all this straight with the different transformations and coordinate systems so a second set of eyes would be useful.

ivalaginja:
Looking at this asap.

@ehpor might also be useful to get your comments on this.

steigersg:
@ivalaginja and @RemiSoummer I removed the transform offset function, tested it on HiCAT and everything seems to be operating as expected. 

In testing I realized that we also forgot the gain property and setter in the camera base class so I added that as well

To my knowledge this PR should now be good to go and once it is merged I will make a new branch with just the transform offset functionality.

pyproject.toml
ivalaginja, line 1: I think we're missing the v2 entry points for the allied vision, dummy and flir cameras.

catkit2/services/allied_vision_camera_v2/allied_vision_camera_v2.py
ivalaginja, line 1: This v2 class will need implementation eventually, but this is not needed to successfully merge this PR.

ivalaginja (commented):
Minor comments that can be easily addressed, looks good to me. I opened https://github.com/spacetelescope/catkit2/issues/449 as a follow up.

Question: Should we do service doc entries for the camera base class and the derived classes?

docs/services/camera.rst
```diff
@@ -0,0 +1,99 @@
...
+
+- ``width``: The width of the ROI in pixels
+- ``height``: The height of the ROI in pixels
+- ``offset_x``: The x-offset of the top-left corner of the ROI on the sensor
+- ``offset_y``: The y-offset of the top-left corner of the ROI on the sensor
+
+If width and height are not specified, they default to the full sensor size.
+If offset_x and offset_y are not specified, they default to (0, 0).
+
+All ROI coordinates are expressed in the **camera coordinate system**. With issue #449, we wand to change this
```
steigersg:
```suggestion
All ROI coordinates are expressed in the **camera coordinate system**. With issue #449, we want to change this
```

docs/services/camera.rst
```diff
@@ -0,0 +1,99 @@
...
+      offset_x: 0                       # Defaults to 0
+      offset_y: 0                       # Defaults to 0
+
+      # Exposure and gain configuration
+      exposure_time: 1000               # Exposure time in microseconds. Defaults to 1000
+      gain: 0                           # Gain setting. Defaults to 0
+
+      # Frame transformation configuration (all optional)
+      rot90: false                      # Rotate image by 90 degrees. Defaults to false
+      flip_x: false                     # Flip image horizontally (vertical flip). Defaults to false
```
steigersg:
```suggestion
      flip_x: false                     # Flip image along x-axis (vertical flip). Defaults to false
```
I think using horizontally here was confusing

docs/services/camera.rst
```diff
@@ -0,0 +1,99 @@
...
+      flip_y: false                     # Flip image vertically (horizontal flip). Defaults to false
```
steigersg:
```suggestion
      flip_y: false                     # Flip image along y-axis (horizontal flip). Defaults to false
```

docs/services/camera.rst
```diff
@@ -0,0 +1,99 @@
...
+2. Flip X (vertical flip, ``np.flipud``)
+3. Flip Y (horizontal flip, ``np.fliplr``)
+
+**Region of Interest (ROI)**
+
+All cameras support configurable regions of interest (ROI) defined by:
+
+- ``width``: The width of the ROI in pixels
+- ``height``: The height of the ROI in pixels
+- ``offset_x``: The x-offset of the top-left corner of the ROI on the sensor
```
steigersg: is "top-left" correct? I'm just not positive

ivalaginja (approved):
Went through this in real time with @steigersg and @RemiSoummer just now, good to go.

RemiSoummer (approved):
Went over this with @ivalaginja and @steigersg  and this looks good

PR #132, opened 2023-10-26
Hardcode flip for science camera as a hack

ehpor:
A very temporary hack until #124 is fixed properly. This is necessary due to hardware changes on HiCAT. It will be removed right after #124 is fixed.

Edit: reverted in https://github.com/spacetelescope/catkit2/pull/451


raphaelpclt (approved):
Tested on hardware

Issue #123, opened 2023-10-13
Refactor BostonDM to only have controlled actuator on data streams


ehpor:
This would remove one of three formats for DM commands. All current DM command formats are:

1. The BMC command (shape for HiCAT is 2048).
2. The active/controlled actuators command (shape for HiCAT is 1904). (edit by @ivalaginja, see https://github.com/spacetelescope/catkit2/pull/156#issuecomment-2017632866)
3. The DM maps (shape for HiCAT is 2x 34x34).

After this change, only the last two would remain. This would make it simpler to distinguish between them, since one of them is 2D and the other is 1D. We use format 2 extensively for WFC and format 3 for plotting/display. However, format 1 is only used for communicating with the DM. This could be internalized inside the BMC DM service, being invisible to the outside. The conversion between active/controlled  actuators and BMC command can be done purely based on data. This would also make it easier for the THD2 team to use our exact service for their BMC DMs.

ehpor:
To formalize nomenclature, we could rename format 2 "commands" and format 3 "maps".

Additionally, we can make the DM maps 3D, with the active/controlled  actuators for each map (defined by the dm_mask, also 3D) being concatenated to form the command. This would make the proxy fully data-driven as well.

ivalaginja:
@ehpor I think this was fully addressed with the new DM/BMC (base) classes, right?

ivalaginja:
@ehpor one more ping to check whether we can close this issue.

ivalaginja:
@ehpor and another ping ^^

ivalaginja:
@ehpor this is done, right?

ehpor:
Yes.

PR #156, opened 2024-02-05
Add DM base class.

ehpor:
- [x] Add ServiceProxy.
- [x] Add logic for device actuator map.
- [x] Worry about quantization. Where should that be handled?
- [x] Add startup map logic (partially already there, but not in the service proxy code).


ivalaginja:
Hey @ehpor, I am picking this up as we're at a point on the THD2 implementation that we need the DMs next. I wanted to clarify a couple of points before I put my hand on this (planning a delta PR in case you want to touch this branch here).

In particular, I wanted to ask about the controlled actuators mask. I remember us talking in person about:  
(a) the actuators that a BMC DM can control in principle, which on HiCAT is 952  
(b) the actuators of the DM that the user chooses to control, which can be a subset of the actuators of (a)  

What did we settle on in terms of nomenclature? In https://github.com/spacetelescope/catkit2/issues/123#issue-1942589716, it seems like number (2) in that issue refers to (a) here, but in your PR you use the term "controlled actuators" for this it seems. I know that the concept for (b) is not implemented yet, and you have it on the todo list of this PR, I just wanted to find you in the same place about this before I move on, otherwise it will get very confusing ^^

When referring to (a), do we call those "active" or "controlled" actuators from now on? I like the "controlled" like in your code - if you confirm this, I will make sure to update that catkit2 issue to minimize confusion.

**Edit: @ehpor and I decided together that (a) will be called "device actuators" and (b) can be addressed synonymously as "active" or "controlled" actuators.**

ivalaginja:
I went ahead and updated the comment in https://github.com/spacetelescope/catkit2/issues/123#issue-1942589716 and the one that follows to use "controlled actuators" like in this PR. ~This means that the todo item in the PR description remains as implementation of "active actuators", which are a subset of the controlled actuators.~

See prior comment.

ivalaginja:
@ehpor general question: is this supposed to be a base class for BMC DMs or for DMs in general? I was wondering how to implement the `send_surface()` command, or to leave it unimplemented.

ehpor:
For DMs in general. It's the channels and controlled actuator map that it should abstract away.

ivalaginja:
@ehpor another question: do I interpret all your comments correctly, especially in https://github.com/spacetelescope/catkit2/issues/123, to say that the new service will be usable per each individual DM (e.g. for THD2) as well as per DM pair (like on HiCAT) since all properties are data-driven, but the new proxy will be per *pair* of DM, like it has been done so far as the proxy can be made data-driven as well?

ivalaginja:
> @ehpor another question: do I interpret all your comments correctly, especially in #123, to say that the new service will be usable per each individual DM (e.g. for THD2) as well as per DM pair (like on HiCAT) since all properties are data-driven, but the new proxy will be per _pair_ of DM, like it has been done so far as the proxy can be made data-driven as well?

I don't actually think this would work. I tried to implement the conversion from the DM command to BMC (hardware) command (see https://github.com/spacetelescope/catkit2/issues/123#issuecomment-1762166021) in the last commit (5b5b83a1b2d4c150aae0e694000d51fad961764e - "Convert DM command to BMC command in send_surface()"), but this makes a hard assumption about all controlled actuators being in the beginning of the hardware command, without any gaps in the array entries. I currently have no idea how to make this work for a pair of DMs - if this was ever the intention.

ivalaginja:
I made this work in my delta-PR in https://github.com/spacetelescope/catkit2/pull/182 for one of the DMs on THD2 (**only in simulated mode**). However, many questions remain.

- Currently, I am using the flat map and gain map as DM commands (no. 2 in https://github.com/spacetelescope/catkit2/issues/123#issuecomment-1762166021). If we keep it this way, shouldn't they be moved out into the base class?
- If they stay in the BMC class, should I assume we import a file that is already in the hardware command format (no.1 in https://github.com/spacetelescope/catkit2/issues/123#issuecomment-1762166021) and adjust `send_surface()` accordingly?
- Is the BMC service supposed to support the control of a pair of DMs, or should this happen exclusively through a proxy?
- Is it sufficient to read the BMC hardware command from the device in the BMC service in order to correctly to the conversion of a DM command to the BMC hardware command?

ivalaginja:
@ehpor what I did in the meantime is:
- Move the discretization into the concrete service since there is no unique way to discretize the DM voltages/surface
- Discretize the total surface always as well
- Tried to implement a way to convert to hardware commands that is agnostic to the number of DMs controlled by a single service.

The last point is hard because I have to make assumptions on the type and format of the controlled actuator mask. I remember you talking about stacking those masks into a 3D array, but that doesn't seem possible if the concerned DMs do not have the same shapes. Shall I just stack them into a list instead? This would mean some changes to the proxy, and associated parameters like for example num_actuators and dm_shape.

ivalaginja:
@ehpor I just realized you said to concatenate the actuator masks - I suppose into a ravelled 1D array, like we already do with the flatmap and gain map. Is that what you meant?

ivalaginja:
Notes from chat at SPIE with @ehpor:
1. Each DM works on a separate service and separate proxy
2. Base class should read starting index of hardware command from service config for each DM, array length of the controlled actuators is inferred from the mask read in also from config --> this assumes no gaps in hardware command, exception could be e.g. IrisAO, but not the case for HiCAT right now
3. dm shape is the literal geometrical shape and can be inferred from the shape of the controlled actuator mask
4. number of actuators is the sum of the device actuator mask
5. All input files should be 2D files so they remain human readable - but the attributes in the service should be 3D, always, to indicate number of DMs
6. Other types of DMs like IrisAO or TT/DAC remain with 1D properties since the geometry is not squares
7. Controlled actuator mask is always 3D, where 0th index is number of DMs where everything else is whatever shape of the DM --> this only works if you have multiple DMs controlled by the same service, which only works if they all have the same shape (e.g. multiple DACs) and all controlled with the same driver (and also same channels)
8. Change nomenclature to "device actuators" for physically existing actuators (visual help in fits file) --> this makes it ok to use "controlled" or "active" actuators interchangeably for any subset of the device actuators

ivalaginja:
I think I am slowly getting there. The main question I am trying to figure out right now is: What command/map format do we want the proxy method `apply_shape()` to be able to accept? See this inline comment ffffff for this (@ehpor):
https://github.com/spacetelescope/catkit2/blob/b1aa19e54854235095812808e2e10d8bc8bd0c16/catkit2/testbed/proxies/deformable_mirror.py#L96

The other thing I will have to check for is whether all of this is compatible with DMs without maps that contain the DM geometry, like a DAC or an IrisAO (see point 6 in [this comment](https://github.com/spacetelescope/catkit2/pull/156#issuecomment-2180108436) further above).

And for now, some notes on the status quo:

- The `device_command_index` in the service configuration can be _undefined_ , an _integer_ or a _list_. In the first case, zero will be assumed. In the second case, only one device is controlled with the service, and in the third case we have several devices of the same number of actuators controlled with the same service and their indices could be different (do we need this)?
- The data streams now hold exclusively (squeezed) 1D DM commands, which means their length is `self.num_actuators * self.num_dms`
- The service has two attributes for the flat map and gain map each: one is a 3D cube of the flat map (gain map) data called `self.flat_map` (`self.gain_map`), the other is a 1D DM command called `self.flat_map_command` (`self.gain_map_command` and `self.gain_map_inv_command`) to be able to operate on other DM commands.
- The service-internal method `send_surface()` exclusively takes a squeezed DM command (the total surface as read from the incoming data frame from the data stream). This means that if DM maps are used to control the DM, they need to be converted to DM commands _in the proxy_.

Open Todos:
- [x] Decide on acceptable input formats of `apply_shape()` in the new proxy.
- [x] Double-check that hardware/device command format is indeed only used in hardware communication and nowhere else.
- [x] Check compatibility with DMs that have non-geometrically encoded DM maps, like a DAC or an IrisAO
- [x] Test, test, test... in simulation initially.

ivalaginja:
@ehpor do you think you could do an initial review of this, to see whether I am on the right path?

ehpor:
@ivalaginja I agree with the decisions you've made so far. 

As for the `apply_shape()`, I'd say there are only two cases:
* (N,) with all device actuators in sequence. This corresponds to the format of the data streams so you don't have do do any formatting.
* (X,*dm_shape) or (*dm_shape) (if X=1, optional). This corresponds to DM maps, which can be converted into the data stream format.

I have not thought about enough on how this converts to DAC and IrisAO.

catkit2/services/bmc_deformable_mirror/bmc_deformable_mirror.py
```diff
@@ -0,0 +1,151 @@
...
+        self.gain_map = fits.getdata(self.gain_map_fname)
+        if self.gain_map.ndim <= 1:
+            raise ValueError(f'The provided gain map for {self.service_id} needs to be at least a 2D array.')
+        elif self.gain_map.ndim == 2:
+            self.gain_map = np.expand_dims(self.gain_map, axis=0)
+        # Convert to DM command
+        self.gain_map_command = self.gain_map[self.device_actuator_mask]
+
+        with np.errstate(divide='ignore', invalid='ignore'):
+            self.gain_map_inv_command = np.ones_like(self.gain_map_command)
```
ehpor: Not sure what's going on here. This immediately gets rewritten in the next line...

catkit2/testbed/proxies/deformable_mirror.py
```diff
@@ -0,0 +1,120 @@
...
+        dm_maps = np.empty_like(self.device_actuator_mask, dtype='float')
+
+        dm_maps[self.device_actuator_mask] = command.reshape(self.dm_shape)  # TODO: Is this going to work?
+
+        return dm_maps
+
+    def apply_shape(self, channel, command):
+        command = np.array(command, copy=False)
+
+        # TODO: with how many potential cases do we want to deal here?
```
ehpor: As said, I want to do away with many command shapes. It should only be (N,) and (X,...) where N is the total number of actuators that this DM controls, and X is the number of DMs and ... is the shape that encompasses all DMs. The first is the format we use for WFC and what we are gonna have on the data streams. The second is just easier to look at / reason about geometry-wise.

catkit2/testbed/proxies/deformable_mirror.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Returns
+        -------
+        array
+            A 3D array with its first dimension giving the number of devices controlled with this service. Second and
+            third dimension are 2D DM maps.
+        """
+        dm_maps = np.empty_like(self.device_actuator_mask, dtype='float')
+
+        dm_maps[self.device_actuator_mask] = command.reshape(self.dm_shape)  # TODO: Is this going to work?
```
ehpor: I think this'll work, yes.

catkit2/testbed/proxies/deformable_mirror.py
```diff
@@ -0,0 +1,120 @@
...
+        command : array
+            A 1D array containing the DM command.
+
+        Returns
+        -------
+        array
+            A 3D array with its first dimension giving the number of devices controlled with this service. Second and
+            third dimension are 2D DM maps.
+        """
+        dm_maps = np.empty_like(self.device_actuator_mask, dtype='float')
```
ehpor: You probably want a `zeros_like()`, to have some control over the unset values.

ehpor (commented):
Very rough review. Not much to say.

Should we have a base service BmcDeformableMirror that then a BmcDeformableMirrorHardware and BmcDeformableMirrorSim derive from? The base service could implement the discretization and the `send_surface()` could just be implemented in the derived classes. Just thinking out loud.

ivalaginja:
@ehpor I implemented your suggestions on command formats and an extra abstraction layer for the BMCs. It works on a **simulated** pair of DMs for THD2, each controlled by each own service.

I suspect some indexing errors still when trying to control a pair of DMs with a single service since this needs to be tested with hicat-package2 to confirm it would be compatible with that. I can do this soon-ish, however, I have currently no way of testing this on hardware.

What's the plan forward apart from the software tests? I will likely start rewriting thd-controls to reflect this refactor since it's our only way forward anyway. A hardware test on our side is tbd and could potentially happen towards the end of August.

ivalaginja:
@ehpor one thing I wasn't sure about, is it ok if there are two service scripts inside of the base folder `base_services`?

ivalaginja:
A quick trial on hicat-package2 shows that this works there notionally. I only tried sending basic commands though. The one thing that broke things is using startup maps, although it is unclear to me whether the issues come form the service of the format of the startup maps in the simulator interface.

A full test would mean to deal with all the inputs to `apply_shape()` everywhere, the correct return of maps after transitioning from `command_to_dm_shapes()` to the new `command_to_dm_maps()`, the correct reformatting of the DM device actuator mask since now it has an extra dimension, and so on. There are also some stuff in the DMs in the optical model that needs to be taken into account, and most gross, the DM viewer.

ivalaginja (approved):
@ehpor tested this today in simulation and we tested it with @ehpor , @raphaelpclt and @RemiSoummer on HiCAT on hardware as well. Ready to go.

PR #163, opened 2024-03-04
Add Hamamatsu camera service

### LINKED ISSUE #162: Add service for Hamamatsu camera
https://www.hamamatsu.com/us/en/product/cameras/driver-software

ivalaginja:
Fixes #162.

Contains a code sneak in which I delete a property from the AV cam service that is not set up or used.

### Todo
- [x] Check library loading
- [x] Test on hardware
- [x] Check dimensions: sensor_width vs. sensor_heigth, width vs. height
- [x] Address all todos in code

### DIFF
--- a/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
+++ b/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
@@ -118,7 +118,7 @@ class HamamatsuCamera(Service):
                              ', must be one of ' +
                              str(list(self.pixel_formats.keys())))
         self.log.info('Using pixel format: %s', self.current_pixel_format)
-        # self.cam.set_pixel_format(self.pixel_formats[self.current_pixel_format])   # TODO
+        self.cam.prop_setvalue(dcam.DCAM_IDPROP.IMAGE_PIXELTYPE, self.pixel_formats[self.current_pixel_format])
 
         # Set device values from config file (set width and height before offsets)
         offset_x = self.config.get('offset_x', 0)
@@ -258,7 +258,7 @@ class HamamatsuCamera(Service):
         float:
             The temperature of the camera in degrees Celsius.
         """
-        return self.cam.DeviceTemperature.get()   # TODO
+        return self.cam.prop_getvalue(dcam.DCAM_IDPROP.SENSORTEMPERATURE)
 
     @property
     def exposure_time(self):
@@ -269,13 +269,13 @@ class HamamatsuCamera(Service):
 
         Returns:
         --------
-        int:
-            The exposure time in microseconds.   # TODO
+        float:
+            The exposure time in seconds.
         """
-        return self.cam.ExposureTime.get()   # TODO
+        return self.cam.prop_getvalue(dcam.DCAM_IDPROP.EXPOSURETIME)
 
     @exposure_time.setter
-    def exposure_time(self, exposure_time: int):
+    def exposure_time(self, exposure_time: float):
         """
         Set the exposure time in microseconds.
 
@@ -283,10 +283,10 @@ class HamamatsuCamera(Service):
 
         Parameters
         ----------
-        exposure_time : int
-            The exposure time in microseconds.   # TODO
+        exposure_time : float
+            The exposure time in seconds.
         """
-        self.cam.ExposureTime.set(exposure_time)   # TODO
+        self.cam.prop_setvalue(dcam.DCAM_IDPROP.EXPOSURETIME, exposure_time)
 
     @property
     def gain(self):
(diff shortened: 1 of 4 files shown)


ivalaginja:
For once a camera service that was simple and went smoothly. Tested on hardware on a Hamamatsu ORCA-Quest C15550-20UP as the `science_camera` with a 400 x 400 px ROI, and everything works as expected:

Screenshot 2024-04-18 at 13 42 07

ivalaginja:
@ehpor note how this is the first camera in catkit2 that uses seconds as exposure time units, instead of microseconds as for the other cameras. I have formulated an issue around this here: https://github.com/spacetelescope/catkit2/issues/181

Edit: the service has been changed since to use microseconds as exposure time as well.

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -0,0 +1,494 @@
...
+            while self.should_be_acquiring.is_set() and not self.should_shut_down:
+                if self.cam.lasterr().is_timeout():
+                    self.log.warning('Timeout while waiting for frame')
+                if self.cam.wait_capevent_frameready(timeout_millisec) is False:
+                    raise RuntimeError(f'Dcam.wait_capevent_frameready({timeout_millisec}) fails with error {self.cam.lasterr()}')
+
+                if self.cam.buf_getframedata(-2) is False:
+                    # If there is no second to last frame, it means we just acquired the first frame.
+                    # In this case, drop the current frame since it always contains systematic errors.
+                    break
```
ivalaginja: While testing in the lab, we learned that the first frame of each acquisition contains very obvious systematic errors, so it is unusable. It is probably rare that anyone would use the very first frame from a data stream, but it seemed more correct to me to never submit that faulty frame to the data stream in the first place.

I am not actually sure how to test this code block though... if you have any pointers @ehpor, please let me know.

@erinpougheon could I ask you to post a screenshot of the faulty frame here in the PR, for future reference?

ehpor: Question: does this faulty frame reappear when you change ROI, exposure time, gain or other settings?

Yes, I'd never submit that frame to the data stream. I'm not sure what you mean by "test". Since this is hardware-only behaviour, you can't do unit tests unless you go for hardware-in-the-loop, which is a hassle to automate. If you mean, just testing that this behaves as expected, I'd put an info log message instead of just breaking.

ivalaginja: Very good questions. I'll pass them on to @erinpougheon to test in the lab when she's back on Wednesday. FYI @johanmazoyer 

I meant indeed to just test whether it behaves as expected. I'll do that.

erinpougheon: Faulty frame
Here is a screenshot of the faulty frame.

ivalaginja:
> Question: does this faulty frame reappear when you change ROI, exposure time, gain or other settings?

So it turns out that we haven't paid much attention to the proper way of property setting with this camera - or the Allied Vision cameras. Some of the properties need the camera acquisition to be temporarily stopped, so I opened an issue for this for the AV service here: https://github.com/spacetelescope/catkit2/issues/183

For the Hamamatsu camera, only the exposure time can be changed while the camera is acquiring, the other properties need a temporary stop of the acquisition as well. Since that retriggers the acquisition loop, yes the faulty frame would appear again, but it would also be automatically discarded again in the current implementation. As for resetting the exposure time, I don't actually know how to catch the exact first frame after resetting it to figure out whether it's fine or not...

We only have our loaner camera for one more day and we won't be able to implement and test this before we have to send it back. We won't get our own camera for a while. My idea would be to open an issue an put a rework of the properties that need an acquisition halt into it (https://github.com/spacetelescope/catkit2/issues/194), akin to the issue for the Allied Vision cameras. And to open a separate issue (#195) to figure out how to test the first frame upon an exposure time change to decide whether it's fine or whether we need to throw it away too.

What do you say?

ivalaginja:
The tests on GitHub are failing for MacOS for some reason, but they are successful when I run them locally:

Screenshot 2024-04-24 at 15 01 21

ehpor:
> What do you say?

Yes, I'm fine with merging this early, with potentially incorrect first-frame handling and some other issues to be fixed in a later PR.

ehpor:
Let me know when this is ready for review, since you're still changing things.

ivalaginja:
@ehpor I think this is actually ready for review now, unless I forgot anything that we talked about.

ehpor (approved):
LGTM.

PR #166, opened 2024-03-14
Refactor service architecture for better visibility.

a-sevin:
This is a proposal for a new service architecture to gain visibility.
It should not break the previous configuration of the bench.


a-sevin:
Closed, superseded by https://github.com/spacetelescope/catkit2/pull/251.

PR #170, opened 2024-03-25
Add services for Thorlabs motors using thorlabs-apt

### LINKED ISSUE #169: Add services for Thorlabs TDC001 and KDC101 cube motors
Summary of the dev process and Python library choice:

Initially, we thought we would be using [https://gitlab.com/ptapping/thorlabs-apt-device](https://gitlab.com/ptapping/thorlabs-apt-device) since it's very clean and the API is easy to use. However, the service using this library we ran into some issues as described here https://github.com/spacetelescope/catkit2/pull/170#issuecomment-2047456686. The original service using this API can be found in the history of https://github.com/spacetelescope/catkit2/pull/170 (the respective service has been deleted).

We then moved on to use [https://pypi.org/project/thorlabs-apt/](https://pypi.org/project/thorlabs-apt/) instead, which is the latest service in https://github.com/spacetelescope/catkit2/pull/170. This works fine when running a single service, both a TDC as well as a KDC, however, we are not able to run more than one motor service like this at a time, see https://github.com/spacetelescope/catkit2/pull/170.

At this point we went back to try to resolve the issues we had been having with [https://gitlab.com/ptapping/thorlabs-apt-device](https://gitlab.com/ptapping/thorlabs-apt-device). In short, there is no way to access the serial number of a device until the serial port was actually opened and the connection established. And even so, the returned serial numbers seem to be garbage, so there seems to be no way to identify motors by serial number. We would have to keep track of the respective (virtual) COM ports and that would likely get too tedious with 10+ motors.

At this point, @a-sevin suggested to install the [Kinesis 64-Bit Software for 64-Bit Windows](https://www.thorlabs.com/software_pages/viewsoftwarepage.cfm?code=Motion_Control#) and use their official Python library, https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python, which had also been suggested here: https://github.com/spacetelescope/catkit2/pull/170#issuecomment-2047879545.

After the Kinesis installation, @a-sevin then found the following, cleaner examples of how to implement this, both under an MIT license:
https://github.com/MSLNZ/msl-equipment/blob/main/msl/examples/equipment/thorlabs/kdc101.py
https://github.com/aquilesC/experimentor/blob/main/experimentor/drivers/thorlabs/tdc001.py

### Final choice

Using bits of the official vendor Python library https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python with an installation of the Kinesis software, see PR #200.

ivalaginja:
Fixes #169.

Uses https://github.com/qpit/thorlabs_apt.

Edit: I still didn't quite give up on using [thorlabs-apt-device](https://gitlab.com/ptapping/thorlabs-apt-device/) since it has proven really hard to make more than one motor work. I am in conversation with the package author here: https://gitlab.com/ptapping/thorlabs-apt-device/-/merge_requests/3

### Todo before review
- [x] Write service proxy
- [x] Figure out motor units
- [x] Write proxy
- [x] Test on hardware for KDC101 motor
- [x] Test on hardware for TDC001 motor
- [ ] Test on 2 motors at the same time
- [ ] Test with a Hub
- [ ] Add simulated service


erinpougheon:
(minor Johan's edits)

We chose to a new service using "**thorlabs_apt**" python package instead of "**thorlabs_apt_device**" as initially started.
https://pypi.org/project/thorlabs-apt/

We changed because we had various problems with thorlabs_apt_device 
- we could not recover a good serial number of the cube 
- therfore we could only access the cube using an identification by the serial port (COM1, COM2, COM3, etc) of the computer / cube on which the motor is connected, which mean that we are dependent on the order of connection. 

On the other hand thorlabs_apt is using the Thorlabs APT DLL (and is therefore much simpler) and allow access to the cube directly using the SN. Thorlabs_apt also provides the motor units, using directly real-life units (mm or deg). It use the same framework form to T and K cube. On paper, the **thorlabs_apt**" python package seems quite old and unfrequently updated (not updated since 2019) compared to **thorlabs_apt_device**. However, the fact that this is only a python wrapper of the DLL makes it a lot simpler et more stable (as long as Thorlabs are not updating the DLL ofr our motors). All tests with this package were performed in python3.9. 

I did some tests (without the service, just thorlabs_apt) to see how the motor reacted to thorlabs_apt's methods and everything worked well, as expected. We may have to do some minor changes in the code in the thorlabs_apt. Indeed, it currently requires to copy/paste the APT.dll file, and we would like the code to go and find it on its own. 

I performed tests : the new service open successfully on THD server. But I can't perform tests to move motors (problems with the proxy. I thought we would't need one but it seems that we will need one). I can't fix the problem or code the proxy properly to make tests work.

johanmazoyer:
NB : On another note, we found that, on the Thorlabs's GitHub, they are developing python code for the K-Cube (the T-Cube is not coded yet). This is very recent (K cube was added only a few months ago) and not very regular dvp. They also use the DLL. Their code isn't ready so we can't use it but it might be useful in the future. 

https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
+from catkit2.testbed.service import Service
+
+import thorlabs_apt as apt
+import numpy as np
+
+
+class ThorlabsCubes(Service):
```
ivalaginja: The service class and file name, as well as the directory they are in all need to have the same name for things to work smoothly in catkit2, while abiding to the camel case vs. underscore notations - all of which you already did. I would also make the name very explicit by adding "motor" to it. One option would be to use the name "Thorlabs Cube Motor", which would mean `ThorlabsCubeMotor` for the class, `thorlabs_cube_motor.py` for the file and `thorlabs_cube_motor` for the directory.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
+from catkit2.testbed.service import Service
+
+import thorlabs_apt as apt
+import numpy as np
+
+
+class ThorlabsCubes(Service):
+
+    def __init__(self):
+        super().__init__('thorlabs_cubes')
```
ivalaginja: This would then also have to be changed to `thorlabs_cube_motor` if you choose my suggestion above, otherwise adjust to whatever you choose to use.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __init__(self):
+        super().__init__('thorlabs_cubes')
+
+        self.cube_serial_number = self.config['serial_number']
+
+        self.command = self.make_data_stream('command', 'float64', [1], 20)
+        self.current_position = self.make_data_stream('current_position', 'float64', [1], 20)
+
+        # Initialization of the parameters
```
ivalaginja: Either make this comment more explicit or delete it - setting a bunch of variables to `None` is very obviously an initialization (commenting code is always good, especially while developing, but excessive comments that are redundant just clutter the code).

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+        self.motor = None
+        self.cube_model = None
+        self.motor_unit = None
+        self.motor_max = None
+        self.motor_min = None
+
+    def open(self):
+        """
+            Open the motor and cube.
```
ivalaginja: Aren't the motor and the cube the same thing, at least for the API? If yes, just cut it to say "Open connection to the motor.", if not, explain.
Also, there is one indentation level too much before the text, it should align with the quotation marks in the line above.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+
+            This will also check if the configured characteristics are matching the one the cube have.
```
ivalaginja: One indentation level too much.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        self.motor = apt.Motor(self.cube_serial_number)
+
+        # Retrieve min, max, unit and model of the motor and cube
```
ivalaginja: Since you want to do a comparison between these parameters as they come from the device itself with what we feed it from the configfile, be more explicit in this comment, e.g.: "Retrieve the min, max, unit and model from the connected device."

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+        min = self.motor.get_stage_axis_info()[0]
+        max = self.motor.get_stage_axis_info()[1]
+        unit = ''
+        model = self.motor.hardware_info[0][2:-1]
+        if self.motor.get_stage_axis_info()[2] == 1:
+            unit = 'mm'
+        if self.motor.get_stage_axis_info()[2] == 2:
+            unit = 'deg'
+
+        # Config min, max, unit and model of the motor and cube
```
ivalaginja: And then here make it more explicit towards the other end, e.g.: "Read min, max, unit and model from service configuration."

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+            Open the motor and cube.
+
+            This will also check if the configured characteristics are matching the one the cube have.
+        """
+        self.motor = apt.Motor(self.cube_serial_number)
+
+        # Retrieve min, max, unit and model of the motor and cube
+        min = self.motor.get_stage_axis_info()[0]
+        max = self.motor.get_stage_axis_info()[1]
+        unit = ''
```
ivalaginja: It might be better to set `unit = None` here so that a simple type check tells us whether it was able to set a unit or not. Unless there is a specific reason why you want this to be an empty string.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+        if self.motor.get_stage_axis_info()[2] == 2:
+            unit = 'deg'
+
+        # Config min, max, unit and model of the motor and cube
+        self.min_pos = self.config['motor_min_pos']
+        self.max_pos = self.config['motor_max_pos']
+        self.unit = self.config['motor_unit']
+        self.cube_model = self.config['cube_model']
+
+        # Compare
```
ivalaginja: Be more explicit about what you are comparing, or delete this comment. If you decide to keep it, one option would be something like: "Compare the device parameters to the service configuration."

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+        try:
+            self.min_pos == min
+            self.max_pos == max
+            self.unit == unit
+            self.cube_model == model
+        except Exception:
+            raise
```
ivalaginja: Just because the equality checks return `False` doesn't mean that an exception will be raised, which is what you actually want here.
You could replace the try... except with:
```python
if not config_values == device_values:
   ValueError('Device parameters don't match configuration parameters.')
```
where `config_values` is a list or array of the variables you get from the `self.config` and `device_values` the same but read from the device. Just be careful about comparing lists or arrays with each other, I recommend to test that in an ipython terminal or similar.

Also, this is the type of thing you should test on hardware while developing. Put some garbage into the respective values in the configfile and try to start the service. If you get the error you planted there, everything is fine.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+                frame = self.command.get_next_frame(10)
+            except Exception:
+                # Timed out. This is used to periodically check the shutdown flag.
+                continue
+            self.set_current_position(frame.data[0])
+
+    def close(self):
+        self.motor._cleanup()
+
+    def set_current_position(self, position):
```
ivalaginja: If you have figured out the command units, could you add a docstring here and mention them, so we can look it up easily in the future?

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+            self.sleep(0.1)
+        # Update the current position data stream.
+        self.get_current_position()
+
+    def is_in_motion(self):
+        """Check if the motor is currently moving."""
+        return self.motor.is_in_motion()
+
+if __name__ == '__main__':
+    service = ThorlabsCubes()
```
ivalaginja: Don't forget to update this here when you change the class name.

environment.yml
```diff
@@ -38,5 +38,6 @@ dependencies:
   - flake8
   - pip:
     - dcps
+    - thorlabs-apt-device
```
ivalaginja: You will need to update this to the new package you decided to use. Make sure to check whether it can be installed through conda, in which case you can delete this line here and add your package further above to the list of packages that get automatically installed from conda. If it can only do pip, then you replace it here.

ivalaginja (commented):
@erinpougheon this is looking good! Before we turn to the proxy, I am giving you some comments here to address to clean up the service.

If you can indeed control both the KDC and TCD motors with this service, then say this in the class docstring. Then, delete the directories and files I originally created specifically for the TDC and KDC motors, since we don't be needing those anymore. We will then have to make sure to test the service on both models before requesting an official review.

Also, make sure to update the PR description with a link to the package you decided to use.

ivalaginja:
> I did some tests (without the service, just thorlabs_apt) to see how the motor reacted to thorlabs_apt's methods and everything worked well, as expected. We may have to do some minor changes in the code in the thorlabs_apt. Indeed, it currently requires to copy/paste the APT.dll file, and we would like the code to go and find it on its own.

You should be able to solve this in the same way like for the Thorlabs picomotors, by adding a function that finds the dll for you (click link to see full function, and see how it is used):
https://github.com/spacetelescope/catkit2/blob/0a6108ba57565e51419c6090fc356453ef93eaad/catkit2/services/thorlabs_pm/thorlabs_pm.py#L27

**Edit**: I just saw in the [README of thorlabs_apt](https://github.com/qpit/thorlabs_apt/blob/master/README.md) that this should work just fine if the dll is located under `Windows\System32`, which is probably the default installation location. Can you test whether the code works if you put the dll there? Then we don't need to change the service code and just need to not mess up the installation of APT on the new machine.

ivalaginja:
> NB : On another note, we found that, on the Thorlabs's GitHub, they are developing python code for the K-Cube (the T-Cube is not coded yet). This is very recent (K cube was added only a few months ago) and not very regular dvp. They also use the DLL. Their code isn't ready so we can't use it but it might be useful in the future.
> 
> https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python

@johanmazoyer can you open a low-priority issue on thd-controls to remind us to look into it in the future, like I did for VimbaX? Like that we won't forget about it, even if we don't need it right now.

environment.yml
ivalaginja: I just checked on the GitHub page of the new package and they distribute it neither through conda nor pip, so you'll have to add a GitHub installation like for example for Vimba. This is not ideal but should do it for now.

The more reason for @johanmazoyer to open that issue I mentioned so that we keep track of alternatives and investigate them later.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
```diff
@@ -0,0 +1,108 @@
...
+        self.motor.stop_profiled()
+
+        while self.is_in_motion():
+            self.sleep(0.1)
+        # Update the current position data stream.
+        self.get_current_position()
+
+    def is_in_motion(self):
+        """Check if the motor is currently moving."""
+        return self.motor.is_in_motion()
```
ivalaginja: What type is this returning?

ivalaginja:
@erinpougheon please don't forget to open the related PR in thd-controls as well asap! That's the branch you're  using to work in this catkit2 PR, so we'll need both open to be able to get them to work and review them.

environment.yml
johanmazoyer: I think it exists on pip ?  https://pypi.org/project/thorlabs-apt/

ivalaginja: Oh thank goodness. It's not listed on the GitHub page so I haven't seen it. But I just saw you literally posted a link to it somewhere else too.

environment.yml
```diff
@@ -38,5 +38,6 @@ dependencies:
   - flake8
   - pip:
     - dcps
+    - thorlabs_apt
```
ivalaginja: Careful, the installation uses a hyphen and not an underscore.

catkit2/services/thorlabs_cube_motor/thorlabs_cube_motor.py
```diff
@@ -23,36 +22,31 @@ def __init__(self):
...
-        try:
-            self.min_pos == min
-            self.max_pos == max
-            self.unit == unit
-            self.cube_model == model
-        except Exception:
-            raise
+        # Compare the device parameters to the service configuration.
+        if not (self.min_pos == min and self.max_pos == max and self.unit == unit and self.cube_model == model):
+            ValueError("Device parameters don't match configuration parameters.")
```
ivalaginja: An exception needs to be raised, you will need to change it to look like this (I had given you a wrong example in my original review comment, sorry about that):
```python
raise ValueError("Device parameters don't match configuration parameters.")
```

catkit2/services/thorlabs_cube_motor/thorlabs_cube_motor.py
```diff
@@ -23,36 +22,31 @@ def __init__(self):
 
     def open(self):
         """
-            Open the motor and cube.
+        Open connexion to the motor.
```
ivalaginja: "connection" in English :)

ivalaginja (commented):
@erinpougheon and I tested the service in the lab today and we got both a KDC and a TDC motor to move with it. There is an issue with the main loop that needs fixing since the motor closes after the first command is sent, and there is a problem in the close() because it says that the motor method `_cleanup()` does not exist. See related issue on thorlabs_apt: https://github.com/qpit/thorlabs_apt/issues/6

This PR will be put on hold for about a week since we had other priorities pop up in the lab until then.

catkit2/services/thorlabs_cubes/thorlabs_cubes.py
ivalaginja: It looks like it's a bool.

catkit2/services/thorlabs_cube_motor/thorlabs_cube_motor.py
```diff
@@ -0,0 +1,113 @@
...
+                # Get an update for the motor position.
+                frame = self.command.get_next_frame(10)
+                # Set the current position if a new command has arrived.
+                self.set_current_position(frame.data[0])
+            except Exception:
+                # Timed out. This is used to periodically check the shutdown flag.
+                continue
+
+    def close(self):
+        apt._cleanup()
```
ivalaginja: I am unsure of this behavior when we have several motor services running. Also, do we need to call the init on the apt library? Might be a follow-up to keep in mind, and keep this PR just to the development of a single service. See also https://github.com/qpit/thorlabs_apt/issues/6

Maybe the solution would be to create a service that handles a group of these motors, like the Newport XPS: https://github.com/spacetelescope/catkit2/blob/develop/catkit2/services/newport_xps_q8/newport_xps_q8.py

@ehpor any thoughts? I haven't tested the service in the current PR for more than one motor running at once yet, maybe I should do that first?

catkit2/services/thorlabs_cube_motor/thorlabs_cube_motor.py
```diff
@@ -0,0 +1,113 @@
...
+            unit = 'deg'
+
+        # Read min, max, unit and model from service configuration.
+        self.min_pos = self.config['motor_min_pos']
+        self.max_pos = self.config['motor_max_pos']
+        self.unit = self.config['motor_unit']
+        self.cube_model = self.config['cube_model']
+
+        # Compare the device parameters to the service configuration.
+        if not (self.min_pos == min and self.max_pos == max and self.unit == unit and self.cube_model == model):
```
ivalaginja: Can we think of a case where we want to constrain a motor movement to a smaller range than what the motor itself allows? In this case, we might want to replace the strict equality checks with a smaller or equal for the max, and a larger or equal for the min.

ivalaginja: I implemented it like that now, using the min and max values from the service configuration as a safety range, which has to lie within the min and max as read from the device.

catkit2/services/thorlabs_cube_motor/thorlabs_cube_motor.py
ehpor: This differs from implementation to implementation. Some devices require having only a single device loaded at the same time, others allow for multiple. If the documentation doesn't tell you, trying it out is the only way to know.

erinpougheon:
I ran some tests today.
First, I tested the service and proxy with a TDC001, and after some fixes, everything works as expected. The service opened successfully, and then I was able to move the motor thanks to both service and proxy's commands (move_absolute, move_relative, home()). Everything worked as expected, the motor performed the wanted moves and no error was raised. 
Then, I did the exact same tests with a KDC101, and same, everything worked as expected. 

Then, I tryed the next step in our tests: connecting two motor to the computer, to have two different instances of the service running at the same time (before testing to connect a hub with several motors). 
But it wasn't working. The first service's instance start successfully, but when I try to start the second one I have an error (cf screenshot). I tryed with reverse order (start the other motor first, and then starting in second the one I started first previously) but I have exactly the same error, the second service's instance doesn't start. 

image

ivalaginja:
Thanks for the tests! I did some digging and tried to connect to both motors without using the service, and that actually worked just fine:
Screenshot 2024-05-03 at 17 36 06

So I am not sure why we cannot spin up a second service in parallel to the first one, especially since it fails upon initialization of the second device, and not somewhere in loading or initializing the library.

If you have any ideas at all @ehpor, I am a taker. Or if you have a moment to look into this by any chance @a-sevin.

ivalaginja:
I now printed the serial number in the catkit2 service just before calling `apt.Motor(self.serial_number)` in the `open()`. The returned serial numbers are correct but it still fails with the same error message. I really have no clue what is happening.

ivalaginja:
I did another bunch of tests and it's really about having the two services run at the same time. It doesn't matter which one I turn on first and shut down, what sequence. They just can't be running at the same time for some reason.

ivalaginja:
I now also tried using `self.motor.move_to(position, blocking=False)` instead of `blocking=True` since I am not sure what everything is blocked while the motor is moving with that keyword, and whether that causes the problem. It does not change anything. I will still remove the keyword since the default is False.

ivalaginja:
Retiring this PR in favor of using the Thorlabs Kinesis software instead of APT (implemented in PR #200), see https://github.com/spacetelescope/catkit2/issues/169#issue-2205418848 for details.

PR #172, opened 2024-04-05
Add services for Ocean Optics spectrometers 

johanmazoyer:
Official API exists in C by Ocean optics called seabreeze: https://www.oceaninsight.com/globalassets/catalog-blocks-and-images/software-downloads-installers/javadocs-api/seabreeze/html/index.html

Good python package wrapping this API seems to be python-seabreeze
https://pypi.org/project/seabreeze/
https://github.com/ap--/python-seabreeze

Address issue in ivalaginja/thd-controls#143

### DIFF
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -27,12 +27,10 @@ class OceanOpticsSpectrometer(Service):
         The model of the spectrometer.
     pixels_number : int
         number of pix in the spectrum.
-    wavelengths : NDArray[numpy.float_]
-        A NDArray with the wavelengths of the spectrometer.
     spectra : DataStream
         A data stream to submit spectra from the spectrometer.
-    should_be_acquiring : threading.Event
-        An event to signal whether the spectrometer should acquire a spectrum.
+    is_saturating : DataStream
+        A data stream to submit whether the spectrometer is currently saturating.
     NUM_FRAMES : int
         The number of frames to allocate for the data streams.
 
@@ -44,14 +42,10 @@ class OceanOpticsSpectrometer(Service):
         The main function of the service.
     close()
         Close the service.
-    start_acquisition()
-        Start the acquisition.
     take_one_spectrum()
         Get a spectrum from the spectrometer.
-    exposure_time()
-        Set the exposure time of the spectrometer.
     '''
-    NUM_FRAMES = 20
+    NUM_FRAMES = 3
 
     def __init__(self):
         '''
@@ -64,13 +58,10 @@ class OceanOpticsSpectrometer(Service):
         self.model = None
         self.pixels_number = None
 
-        self.wavelengths = None
         self.spectra = None
+        self.is_saturating = None
 
         self._exposure_time = None
-
-        self.should_be_acquiring = threading.Event()
-        self.should_be_acquiring.set()
 
     def open(self):
         '''
@@ -95,19 +86,28 @@ class OceanOpticsSpectrometer(Service):
         # exctract attributes
         self.model = self.spectrometer.model
         self.pixels_number = self.spectrometer.pixels
-        self.wavelengths = self.spectrometer.wavelengths()
 
         # Define and set defaut exposure time
         self._exposure_time = self.config.get('exposure_time', 1000)
         self.exposure_time = self._exposure_time
 
         # Create datastreams
-        # Use the full sensor size here to always allocate enough shared memory.
         self.spectra = self.make_data_stream('spectra', 'float32', [self.pixels_number], self.NUM_FRAMES)
 
-        self.make_command('start_acquisition', self.start_acquisition)
-        self.make_command('end_acquisition', self.end_acquisition)
+        self.is_saturating = self.make_data_stream('is_saturating', 'int8', [1], self.NUM_FRAMES)
+        self.is_saturating.submit_data(np.array([0], dtype='int8'))
+
         self.make_command('take_one_spectrum', self.take_one_spectrum)
+
+        # Create properties
+        def make_property_helper(name, read_only=False):
+            if read_only:
+                self.make_property(name, lambda: getattr(self, name))
+            else:
+                self.make_property(name, lambda: getattr(self, name), lambda val: setattr(self, name, val))
+
+        make_property_helper('exposure_time')
+        make_property_helper('wavelengths', read_only=True)
 
     def main(self):
         '''
@@ -116,27 +116,32 @@ class OceanOpticsSpectrometer(Service):
         This function is called when the service is started.
         '''
         while not self.should_shut_down:
-            if self.should_be_acquiring.wait(0.05):
-                self.take_one_spectrum()
-                #self.end_acquisition()
-
-    def start_acquisition(self):
-        '''
-        Start the acquisition loop.
-        '''
-        self.should_be_acquiring.set()
-
-    def end_acquisition(self):
-        '''
-        End the acquisition loop.
-        '''
-        self.should_be_acquiring.clear()
+            self.take_one_spectrum()
 
     def take_one_spectrum(self):
         '''
         Measure one spectrum and submit it to the data stream.
         '''
-        self.spectra.submit_data(np.array(self.spectrometer.intensities(), dtype='float32'))
+        intensities = self.spectrometer.intensities()
+
+        self.spectra.submit_data(np.array(intensities, dtype='float32'))
+
+        if np.max(intensities) ==  self.spectrometer.max_intensity:
+            self.is_saturating.submit_data(np.array([1], dtype='int8'))
+        else:
+            self.is_saturating.submit_data(np.array([0], dtype='int8'))
+
+    @property
+    def wavelengths(self):
+        '''
+        The wavelengths of the spectrometer.
+
+        Returns:
+        --------
+        ndarray of float:
+            wavelengths of the spectrometer.
+        '''
+        return self.spectrometer.wavelengths()
 
     @property
     def exposure_time(self):
@@ -156,6 +161,8 @@ class OceanOpticsSpectrometer(Service):
         Set the exposure time in microseconds.
 
         This property can be used to set the exposure time of the spectrometer.
+        if exposure_time < min allowed exp time: exposure_time =  min allowed exp time
+        if exposure_time > max allowed exp time: exposure_time =  max allowed exp time
 
         Parameters
         ----------
@@ -168,9 +175,10 @@ class OceanOpticsSpectrometer(Service):
             If the exposure time is not in the range of the accepted spectrometer values.
         '''
         int_time_range = self.spectrometer.integration_time_micros_limits
-        if exposure_time < int_time_range[0] or exposure_time > int_time_range[1]:
-            raise ValueError(
-                f"OceanOptics: integration time need to be in [{int_time_range[0]}, {int_time_range[1]}] range (ms)")
+        if exposure_time < int_time_range[0]:
+            exposure_time = int_time_range[0]
+        if exposure_time > int_time_range[1]:
+            exposure_time = int_time_range[1]
 
         self._exposure_time = exposure_time
         self.spectrometer.integration_time_micros(exposure_time)
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -1,7 +1,7 @@
 '''
 This module contains a service for Ocean Optics Spectrometers.
 '''
-
+import numpy as np
 import threading
 
 from seabreeze.spectrometers import Spectrometer
@@ -99,11 +99,15 @@ class OceanOpticsSpectrometer(Service):
 
         # Define and set defaut exposure time
         self._exposure_time = self.config.get('exposure_time', 1000)
-        self.exposure_time(self._exposure_time)
+        self.exposure_time = self._exposure_time
 
         # Create datastreams
         # Use the full sensor size here to always allocate enough shared memory.
         self.spectra = self.make_data_stream('spectra', 'float32', [self.pixels_number], self.NUM_FRAMES)
+
+        self.make_command('start_acquisition', self.start_acquisition)
+        self.make_command('end_acquisition', self.end_acquisition)
+        self.make_command('take_one_spectrum', self.take_one_spectrum)
 
     def main(self):
         '''
@@ -114,7 +118,7 @@ class OceanOpticsSpectrometer(Service):
         while not self.should_shut_down:
             if self.should_be_acquiring.wait(0.05):
                 self.take_one_spectrum()
-                self.end_acquisition()
+                #self.end_acquisition()
 
     def start_acquisition(self):
         '''
@@ -132,7 +136,7 @@ class OceanOpticsSpectrometer(Service):
         '''
         Measure one spectrum and submit it to the data stream.
         '''
-        self.spectra.submit_data(self.spectrometer.intensities(), dtype='float32')
+        self.spectra.submit_data(np.array(self.spectrometer.intensities(), dtype='float32'))
 
     @property
     def exposure_time(self):
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -158,10 +158,10 @@ class OceanOpticsSpectrometer(Service):
 
         This property can be used to set the exposure time of the spectrometer.
 
-        If exposure_time parameter is inferior to the minimum allowed exposure time for this
-        spectrometer (set by ocean optics), it is set to the minimum allowed exposure time.
-        If exposure_time is larger to the maximum allowed exposure time, it is set to the
-        minimum allowed exposure time.
+        If the `exposure_time` parameter is inferior to the minimum allowed exposure time for this
+        spectrometer (set by Ocean Optics), it is set to the minimum allowed exposure time.
+        If the `exposure_time` is larger than the maximum allowed exposure time, it is set to the
+        maximum allowed exposure time.
 
         Parameters
         ----------
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -1,0 +1,155 @@
+'''
+This module contains a service for Ocean Optics Sectrometer
+
+This service is a wrapper around the python seabreeze package
+It provides a simple interface to control the spectro and acquire spectras.
+'''
+
+import seabreeze
+
+seabreeze.use('pyseabreeze')
+
+from seabreeze.spectrometers import Spectrometer as spct
+from catkit2.testbed.service import Service
+
+
+class OceanOpticsSpectro(Service):
+    '''
+    Service for Ocean Optics Spectrometer.
+
+    This service is a wrapper around the python seabreeze package
+    It provides a simple interface to control the spectro and acquire spectras.
+
+    Attributes
+    ----------
+    spectrometer : Spectrometer
+        The spectrometer to control.
+    model : str
+        The model of the spectrometer.
+    pixels_number : int
+        number of pix in the spectras
+    wavelengths : NDArray[numpy.float_]
+        A NDArray with the wl of the spectrometer.
+    spectras : DataStream
+        A data stream to submit the spectras from the spectrometer.
+    NUM_FRAMES : int
+        The number of frames to allocate for the data streams.
+
+    Methods
+    -------
+    open()
+        Open the service.
+    main()
+        The main function of the service.
+    close()
+        Close the service.
+    get_spectra()
+        Get a spectra from the spectrometer.
+    exposure_time()
+        Set the exposure time of the spectrometer.
+
+    '''
+    NUM_FRAMES = 20
+
+    def __init__(self):
+        '''
+        Create a new OceanOpticsSpectro service.
+
+        '''
+        super().__init__('OceanOpticsSpectro')
+
+        self.spectrometer = None
+        self.model = None
+        self.pixels_number = None
+
+        self.wavelengths = None
+        self.spectras = None
+
+    def open(self):
+        '''
+        Open the service.
+
+        This function is called when the service is opened.
+        It initializes the spectrometer and creates the data streams and properties.
+
+        Raises
+        ------
+        RuntimeError
+            If the spectro cannot be found.
+
+        '''
+
+        # Find the spectrometer
+        self.serial_number = str(self.config['serial_number'])
+
+        try:
+            self.spectrometer = spct.from_serial_number(self.serial_number)
+        except:
+            raise RuntimeError(f'Could not find spectrometer with serial number {self.serial_number}')
+
+        # Set the exposure time
+        self.exposure_time = self.config.get('exposure_time', 1000)
+        self.make_property('exposure_time', lambda: getattr(self, 'exposure_time'),
+                           lambda val: setattr(self, 'exposure_time', val))
+
+        # exctract attributes
+        self.model = self.spectrometer.model
+        self.pixels_number = self.spectrometer.pixels
+        self.wavelengths = self.spectrometer.wavelengths()
+
+        # Create datastreams
+        # Use the full sensor size here to always allocate enough shared memory.
+        self.spectras = self.make_data_stream('spectras', 'float32', self.pixels_number, self.NUM_FRAMES)
+
+    def main(self):
+        '''
+        The main function of the service.
+
+        This function is called when the service is started.
+        '''
+        while not self.should_shut_down:
+            spectra = self.get_spectra()
+            self.spectras.submit_data(spectra, dtype='float32')
+
+    def get_spectra(self):
+        return self.spectrometer.intensities()
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectr0.
+
+        Parameters
+        ----------
+        exposure_time : int
+            The exposure time in microseconds.
+        
+        Raises
+        ------
+        ValueError
+            If the exposure time is not in the range of the accepted spectrometer values.
+        '''
+
+        int_time_range = self.spectrometer.integration_time_micros_limits
+        if exposure_time < int_time_range[0] or exposure_time > int_time_range[1]:
+            raise ValueError(
+                f"Ocean Optics: integration time need to be in [{int_time_range[0]}, {int_time_range[1]}] range (ms)")
+
+        self.spectrometer.integration_time_micros(exposure_time)
+
+    def close(self):
+        '''
+        Close the service.
+
+        This function is called when the service is closed.
+        It close the spectro loop and cleans up the data streams.
+        '''
+        self.spectrometer.close()
+        self.spectrometer = None
+
+
+if __name__ == '__main__':
+    service = OceanOpticsSpectro()
+    service.run()
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -116,7 +116,7 @@ class OceanOpticsSpectrometer(Service):
         while not self.should_shut_down:
             if self.should_be_acquiring.wait(0.05):
                 self.get_spectra()
-    
+
     def start_acquisition(self):
         '''
         Start the acquisition loop.
@@ -132,7 +132,6 @@ class OceanOpticsSpectrometer(Service):
         This function ends the acquisition loop.
         '''
         self.should_be_acquiring.clear()
-
 
     def get_spectra(self):
         spectra = self.spectrometer.intensities()
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -2,7 +2,6 @@
 This module contains a service for Ocean Optics Spectrometers.
 '''
 import numpy as np
-import threading
 
 from seabreeze.spectrometers import Spectrometer
 from seabreeze.spectrometers import SeaBreezeError
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -24,6 +24,8 @@ class OceanOpticsSpectrometer(Service):
     ----------
     spectrometer : Spectrometer
         The spectrometer to control.
+    serial_number : str
+        The serial number of the spectrometer.
     model : str
         The model of the spectrometer.
     pixels_number : int
@@ -33,7 +35,7 @@ class OceanOpticsSpectrometer(Service):
     spectras : DataStream
         A data stream to submit the spectras from the spectrometer.
     should_be_acquiring : threading.Event
-        An event to signal whether the spectra should be acquiring a spectrum.
+        An event to signal whether the spectra should acquire a spectrum.
     NUM_FRAMES : int
         The number of frames to allocate for the data streams.
 
@@ -45,7 +47,9 @@ class OceanOpticsSpectrometer(Service):
         The main function of the service.
     close()
         Close the service.
-    get_spectra()
+    start_acquisition()
+        Start the acquisition.
+    take_one_spectrum()
         Get a spectrum from the spectrometer.
     exposure_time()
         Set the exposure time of the spectrometer.
@@ -59,7 +63,8 @@ class OceanOpticsSpectrometer(Service):
 
         '''
         super().__init__('oceanoptics_spectrometer')
-
+        
+        self.serial_number = None
         self.spectrometer = None
         self.model = None
         self.pixels_number = None
@@ -88,7 +93,6 @@ class OceanOpticsSpectrometer(Service):
 
         # Find the spectrometer
         self.serial_number = str(self.config['serial_number'])
-
         try:
             self.spectrometer = spct.from_serial_number(self.serial_number)
         except SeaBreezeError:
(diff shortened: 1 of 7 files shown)


catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
+'''
+This module contains a service for Ocean Optics Sectrometer
```
ivalaginja:
```suggestion
This module contains a service for Ocean Optics Sectrometers.
```

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+
+This service is a wrapper around the python seabreeze package
+It provides a simple interface to control the spectro and acquire spectras.
```
ivalaginja: You have this in the service class, so you don't need it here, it would be good to delete.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+from seabreeze.spectrometers import SeaBreezeError
+
+from catkit2.testbed.service import Service
+
+
+class OceanOpticsSpectrometer(Service):
+    '''
+    Service for Ocean Optics Spectrometer.
+
+    This service is a wrapper around the python seabreeze package
```
ivalaginja:
```suggestion
    This service is a wrapper around the python seabreeze package.
```

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    It provides a simple interface to control the spectro and acquire spectras.
```
ivalaginja:
```suggestion
    It provides a simple interface to control the spectrometer and acquire spectra.
```

Note how "spectra" is the plural form of "spectrum". You'll want to change this everywhere in this file. Also, I think it would be helpful to always write the full word "spectrometer", to avoid any confusion.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
+'''
+This module contains a service for Ocean Optics Sectrometer
+
+This service is a wrapper around the python seabreeze package
+It provides a simple interface to control the spectro and acquire spectras.
+'''
+
+import threading
+
+from seabreeze.spectrometers import Spectrometer as spct
```
ivalaginja: I would drop the alias to keep clarity.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    ----------
+    spectrometer : Spectrometer
+        The spectrometer to control.
+    serial_number : str
+        The serial number of the spectrometer.
+    model : str
+        The model of the spectrometer.
+    pixels_number : int
+        number of pix in the spectras
+    wavelengths : NDArray[numpy.float_]
```
ivalaginja: I think it's just `nadarray` like here:
https://github.com/spacetelescope/catkit2/blob/7663307222cee45655932ded1887ac10e8065d90/catkit2/services/ni_daq/ni_daq.py#L126

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        A NDArray with the wl of the spectrometer.
```
ivalaginja: Spell out wl please.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        The main function of the service.
+    close()
+        Close the service.
+    start_acquisition()
+        Start the acquisition.
+    take_one_spectrum()
+        Get a spectrum from the spectrometer.
+    exposure_time()
+        Set the exposure time of the spectrometer.
+
```
ivalaginja: Blank line needs deletion.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    '''
+    NUM_FRAMES = 20
+
+    def __init__(self):
+        '''
+        Create a new OceanOpticsSpectrometer service.
+
```
ivalaginja: Blank line deletion.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        This function is called when the service is opened.
+        It initializes the spectrometer and creates the data streams and properties.
+
+        Raises
+        ------
+        RuntimeError
+            If the spectro cannot be found.
+
+        '''
+
```
ivalaginja: Both blank lines here need to go.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        while not self.should_shut_down:
+            if self.should_be_acquiring.wait(0.05):
+                self.take_one_spectrum()
+                self.should_be_acquiring.clear()
+
+    def start_acquisition(self):
+        '''
+        Start the acquisition loop.
+        '''
+        self.should_be_acquiring.set()
```
ivalaginja: If you have a start_acquisition() method, you should also have an end_acquisition() method like here:
https://github.com/spacetelescope/catkit2/blob/7663307222cee45655932ded1887ac10e8065d90/catkit2/services/allied_vision_camera/allied_vision_camera.py#L319

and use that in your while loop in main(). *If* you want to have a continuous acquisition mode.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+            The exposure time in microseconds.
+        '''
+        return self._exposure_time
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectr0.
```
ivalaginja: spectr0?

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    def main(self):
+        '''
+        The main function of the service.
+
+        This function is called when the service is started.
+        '''
+        while not self.should_shut_down:
+            if self.should_be_acquiring.wait(0.05):
+                self.take_one_spectrum()
+                self.should_be_acquiring.clear()
```
ivalaginja: It looks to me like this would take exactly one spectrum before stopping the acquisition by clearing the `should_be_acquiring` flag. Is this what you are going for? In this case, where you want to get a spectrum on-command, it might be better to use the variant that for example the DMs are using - you create a thread for a function `monitor_command`. This method monitors the "command" data stream (which needs to be created in the init) inside a while loop. In there, you have a try... except... that tries to get a new frame from the command data stream. If it does, it calls your function `take_one_spectrum()` (the equivalent to `update_dm()` in the DM service), if it doesn't, it just passes and waits until it checks the data stream again.

DM example here, and all the lines that follow, except that you can ditch the for loops as the DM monitors several command channels, and not just one:
https://github.com/spacetelescope/catkit2/blob/7663307222cee45655932ded1887ac10e8065d90/catkit2/services/bmc_dm/bmc_dm.py#L56

If you opt for this, you don't need to bother with any of the start acquisition / end acquisition stuff.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+
+        Returns:
+        --------
+        int:
+            The exposure time in microseconds.
+        '''
+        return self._exposure_time
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
```
ivalaginja, line 155: Are you sure this only takes integers?

ivalaginja (commented):
This is nicely coming into shape! It looks to me like you have made the decision to not have the spectrometer run a continuous acquisition like a camera, and instead to wait for a sent acquisition command like for example the PLP power source or the DM. Is that correct? In this case, I have left you a comment to help you straighten out the `main()` loop. If I am interpreting this wrong though, do let me know and I will point you more in the direction of continuous acquisition loops like the cameras.

Also, could you maybe create a twin PR in https://github.com/ivalaginja/thd-controls with a new service entry for the spectrometer, so that I know which config entries to match and what that looks like?

The rest are just some minor comments that you will need to address before merging anyway and that are not critical to the service.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: apparently, based on the seabreeze pacakge but should be checked with real spectro
Screenshot 2024-04-10 at 18 35 19

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: I copied the type that was in the seabreeze.spectrometers.Spectrometer.wavelengths() function
https://python-seabreeze.readthedocs.io/en/latest/api.html#seabreeze.spectrometers.Spectrometer.wavelengths

Apparently, this is not a typo, there is a difference but I'm not sure to understand. Maybe one is the class the other is the type ? 
https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: Ok, I am not sure to understand. My goal was to give the mean to do both single and loop images, actually, because I am not sure about the use case of the spectro. One observation only could be accessed with spectrometer.take_one_spectrum() or a loop mode with start_acquisition/end_acquisition. Let's talk when you come back, it will be faster than exchangeing over DMs.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+            raise ImportError(f'OceanOptics: Could not find spectrometer with serial number {self.serial_number}')
+
+        # exctract attributes
+        self.model = self.spectrometer.model
+        self.pixels_number = self.spectrometer.pixels
+        self.wavelengths = self.spectrometer.wavelengths()
+
+        # Define and set defaut exposure time
+        self._exposure_time = self.config.get('exposure_time', 1000)
+        self.exposure_time(self._exposure_time)
```
ivalaginja: I think what you could do, but it needs confirmation/a test, is to replace line 70 simply with `self.exposure_time = self.config.get('exposure_time', 1000)`. As long as that is in the init, it sets the default exposure time to this, and then it acts through the getter and setter of the property.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    def main(self):
+        '''
+        The main function of the service.
+
+        This function is called when the service is started.
+        '''
+        while not self.should_shut_down:
+            if self.should_be_acquiring.wait(0.05):
+                self.take_one_spectrum()
+                self.end_acquisition()
```
ivalaginja: It still looks to me like this will take exactly one spectrum and then end the acquisition.

I learned today that the power meter service simplifies this because it doesn't need a super fast acquisition like a camera, see here (and following lines):
https://github.com/spacetelescope/catkit2/blob/98758710dc8696a6d3d3e78aa2a1bf408bf720d5/catkit2/services/thorlabs_pm/thorlabs_pm.py#L20

What it does is that it simply defines an interval in the config entry after which it regularly takes a new measurement. I feel like this would also be the best solution for the spectrometer, as it allows you to ditch the star/end acquisition logic and just have a simple while look in `main()`.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        # exctract attributes
+        self.model = self.spectrometer.model
+        self.pixels_number = self.spectrometer.pixels
+        self.wavelengths = self.spectrometer.wavelengths()
+
+        # Define and set defaut exposure time
+        self._exposure_time = self.config.get('exposure_time', 1000)
+        self.exposure_time(self._exposure_time)
+
+        # Create datastreams
```
ivalaginja, line 91: I think this comment can go since it's a single data stream.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    Attributes
+    ----------
+    spectrometer : Spectrometer
+        The spectrometer to control.
+    serial_number : str
+        The serial number of the spectrometer.
+    model : str
+        The model of the spectrometer.
+    pixels_number : int
+        number of pix in the spectrum.
```
ivalaginja: Please write out "pixels" here.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
ivalaginja: Ok. I am unfamiliar with this so I don't see a problem.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        '''
+        super().__init__('oceanoptics_spectrometer')
+
+        self.serial_number = None
+        self.spectrometer = None
+        self.model = None
+        self.pixels_number = None
+
+        self.wavelengths = None
+        self.spectra = None
```
ivalaginja: We never initialize attributes that hold data streams (maybe with a few exceptions). I have no clue if that's the right way to go, but at this point of time in the project it's probably good to keep things consistent as much as possible.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+    def open(self):
+        '''
+        Open the service.
+
+        This function is called when the service is opened.
+        It initializes the spectrometer and creates the data streams and properties.
+
+        Raises
+        ------
+        RuntimeError
```
ivalaginja: It's an `ImportError` I think

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+        exposure_time : int
+            The exposure time in microseconds.
+
+        Raises
+        ------
+        ValueError
+            If the exposure time is not in the range of the accepted spectrometer values.
+        '''
+        int_time_range = self.spectrometer.integration_time_micros_limits
+        if exposure_time < int_time_range[0] or exposure_time > int_time_range[1]:
```
ivalaginja: You can simplify to:
```python
if not int_time_range[0] < exposure_time < int_time_range[1]:
```

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+            raise ValueError(
+                f"OceanOptics: integration time need to be in [{int_time_range[0]}, {int_time_range[1]}] range (ms)")
```
ivalaginja: "need" --> "needs"

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+
+        self._exposure_time = exposure_time
+        self.spectrometer.integration_time_micros(exposure_time)
+
+    def close(self):
+        '''
+        Close the service.
+
+        This function is called when the service is closed.
+        It close the spectrometer and cleans up the data streams.
```
ivalaginja: "close" --> "closes"

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+            The exposure time in microseconds.
+        '''
+        return self._exposure_time
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectrometer.
```
ivalaginja, line 159: I think this line can also go if you want to reduce some lines. It duplicates the first-line comment.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,187 @@
...
+            self.spectrometer = Spectrometer.from_serial_number(self.serial_number)
+        except SeaBreezeError:
+            raise ImportError(f'OceanOptics: Could not find spectrometer with serial number {self.serial_number}')
+
+        # exctract attributes
+        self.model = self.spectrometer.model
+        self.pixels_number = self.spectrometer.pixels
+        self.wavelengths = self.spectrometer.wavelengths()
+
+        # Define and set defaut exposure time
```
ivalaginja: "l" missing

ivalaginja (commented):
This looks very good to me overall. I would suggest to simplify even more and to just take a measurement each interval of x seconds that is set in the service configuration. We do want a continuous acquisition, but I don't think it needs to be as fast as the camera - one new spectrum every 0.5 seconds or every 1 second should be enough.

This is also a good time to open a new PR on thd-controls to add the service config for the spectrometer there. You need that anyway to be able to test this catkit2 PR.

As for the tests that need to be done on hardware, since the service is very lean, there's only a couple of points:
- Make sure you can open the service and it appears correctly in the service state viewer of the GUI
- Make sure you can correctly update all properties, in this case the exposure time (set a new exposure time and then double-check it registered it by getting the exposure time right after)
- Make sure you can read the spectra from the data stream directly. You might want to turn the different lasers on and off (manually or through code once we connect them to the new machine) to confirm the spectra update correctly on the data stream.
- Once all of that works on hardware, create a documentation page for this service, fill it in (instructions at the bottom of #190) and don't forget to add the new doc page to `index.rst`
- Create a new issue on thd-controls to create a spectrometer viewer; maybe you can already describe a bit what it should look like and what elements it should contain

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: 2 datastream

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: ok

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: ok

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: ok

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: not done anymore

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class OceanOpticsSpectrometer(Service):
+    '''
+    Service for Ocean Optics Spectrometer.
+
+    This service is a wrapper around the python seabreeze package.
+    It provides a simple interface to control the spectrometer and acquire spectra.
+
+    Attributes
```
ivalaginja, line 19: We're missing the `interval` attribute here.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,189 @@
...
+        This function is called when the service is started.
+        '''
+        while not self.should_shut_down:
+            intensities = self.take_one_spectrum()
+            self.spectra.submit_data(np.array(intensities, dtype='float32'))
+            self.sleep(self.interval)
+
+    def take_one_spectrum(self):
+        '''
+        Measure one spectrum and submit it to the data stream.
```
ivalaginja: Since this is not submitting to the `spectra` data stream anymore:
```suggestion
        Measure and return  one spectrum.
```

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,189 @@
...
+        '''
+        intensities = self.spectrometer.intensities()
+
+        if np.max(intensities) ==  self.spectrometer.max_intensity:
+            self.is_saturating.submit_data(np.array([1], dtype='int8'))
+        else:
+            self.is_saturating.submit_data(np.array([0], dtype='int8'))
```
ivalaginja: I have been thinking about it, and I think it might be better if this function really just measures and returns the intensities and to move the saturation check and its data stream submission to the main method too. In this way, the `take_one_spectrum()` can be regarded as a function that does exactly that. It would be a bit cleaner.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,189 @@
...
+        '''
+        return self._exposure_time
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        if exposure_time < min allowed exp time: exposure_time =  min allowed exp time
+        if exposure_time > max allowed exp time: exposure_time =  max allowed exp time
```
ivalaginja: Can you write this out in words? The current docstring essentially just repeats the code and is a little hard to parse and/or unnecessary if the code says almost exactly the same thing. Detail though.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Parameters
+        ----------
+        exposure_time : int
+            The exposure time in microseconds.
+
```
ivalaginja: No blank line here.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,156 @@
...
+import numpy as np
+from catkit2.testbed.service import Service
+
+
+class OceanOpticsSpectrometerSim(Service):
+    '''
+    Service for Ocean Optics Spectrometer.
+
+    This service is a wrapper around the python seabreeze package.
+    It provides a simple interface to control the spectrometer and acquire spectra.
```
ivalaginja: This is not true for the simulated service and should be deleted.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Attributes
+    ----------
+    spectrometer : Spectrometer
+        The spectrometer to control.
+    serial_number : str
+        The serial number of the spectrometer.
+    model : str
```
ivalaginja: All the attributes present in the hardware service should still exist in the simulated service where possible. So please still read the serial number and model from the config in this simulated service as well. You can delete the spectrometer from the Attributes list.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,156 @@
...
+        The model of the spectrometer.
+    pixels_number : int
+        number of pix in the spectrum.
```
ivalaginja: You probably copied this too early ^^ - pleas write out "pixels" here. And capitalize "Number".

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,156 @@
...
+            self.spectra.submit_data(np.array(intensities, dtype='float32'))
+            self.sleep(self.interval)
+
+    def take_one_spectrum(self):
+        '''
+        Measure one spectrum and submit it to the data stream.
+        '''
+        intensities = np.random.random(self.pixels_number) + 1000
+
+        self.is_saturating.submit_data(np.array([0], dtype='int8'))
```
ivalaginja: To follow the comment I left in the hardware service, let's move this data stream submission to the main loop.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,156 @@
...
+    def wavelengths(self):
+        '''
+        The wavelengths of the spectrometer.
+
+        Returns:
+        --------
+        ndarray of float:
+            wavelengths of the spectrometer.
+        '''
+        return np.linspace(350, 1000, num = self.pixels_number)
```
ivalaginja: No spaces around the equal sign.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,156 @@
...
+        return self._exposure_time
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectrometer.
+        if exposure_time < min allowed exp time: exposure_time =  min allowed exp time
+        if exposure_time > max allowed exp time: exposure_time =  max allowed exp time
```
ivalaginja: These lines need to be deleted here.

ivalaginja (commented):
Very cool to see this come alive, the only things left are some touch ups - and the documentation.

I totally forgot about the service docs until I went through the full PR thread again. @johanmazoyer before we can merge this, could you please add a new service doc page for the spectrometer service, fill it in (see instructions in #190), and add it to the doc index file.

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: done

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
johanmazoyer: done

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
johanmazoyer: ok for serial_number. 

Model in not in the yaml, this is a spectrometer attribute that you can get from seabreeze. I thought it could be useful so I created a service attribute for it if someone want to use it wihtout having to parse the documentation of seabreeze package. However, I cannot have it in the simulated service

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,193 @@
...
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectrometer.
+
+        If exposure_time parameter is inferior to the minimum allowed exposure time for this
+        spectrometer (set by ocean optics), it is set to the minimum allowed exposure time.
+        If exposure_time is larger to the maximum allowed exposure time, it is set to the
+        minimum allowed exposure time.
```
ivalaginja:
```suggestion
        If exposure_time is larger than the maximum allowed exposure time, it is set to the
        maximum allowed exposure time.
```

catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,193 @@
...
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectrometer.
+
+        If exposure_time parameter is inferior to the minimum allowed exposure time for this
+        spectrometer (set by ocean optics), it is set to the minimum allowed exposure time.
```
ivalaginja:
```suggestion
        If the `exposure_time` parameter is inferior to the minimum allowed exposure time for this
        spectrometer (set by Ocean Optics), it is set to the minimum allowed exposure time.
```

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
ivalaginja: This still needs to go since it does not use seabreeze. I'd just delete both lines.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,153 @@
...
+    def main(self):
+        '''
+        The main function of the service.
+
+        This function is called when the service is started.
+        '''
+        while not self.should_shut_down:
+            intensities = self.take_one_spectrum(self.pixels_number)
+            self.is_saturating.submit_data(np.array([0], dtype='int8'))
+            self.spectra.submit_data(np.array(intensities, dtype='float32'))
```
ivalaginja: I am really starting to nit-pick but since we're on another review (probs last): Could you invert these two lines? Submit the intensities first (since they're "measured" before), then submit the zero saturation? A blank line before the sleep would also be good.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,153 @@
...
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        Parameters
+        ----------
+        exposure_time : int
+            The exposure time in microseconds.
+
```
ivalaginja: No blank line here.

catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
```diff
@@ -0,0 +1,153 @@
...
+            self.is_saturating.submit_data(np.array([0], dtype='int8'))
+            self.spectra.submit_data(np.array(intensities, dtype='float32'))
+            self.sleep(self.interval)
+
+    def take_one_spectrum(self):
+        '''
+        Measure and return one spectrum.
+        Simulated service: random numpy array.
+        '''
+        return np.random.random(self.pixels_number)
```
ivalaginja: Did you delete the `* 1000` on purpose here? Either way fine by me, just curious.

docs/services/oceanoptics_spectrometer.rst
```diff
@@ -0,0 +1,42 @@
+Ocean Optics Spectrometer
+====================
```
ivalaginja:
```suggestion
========================
```

docs/services/oceanoptics_spectrometer.rst
```diff
@@ -0,0 +1,42 @@
...
+
+This service controls an Ocean Optics Spectrometer. It is a wrapper around the python-seabreeze package.
+The python-seabreeze package needs to be installed first, and the page also explain how to install the
```
ivalaginja, line 5: Very happy you pointed this out because I forgot ^^: We need to add `conda-forge::seabreeze` to the `environment.yml`! And then you can rewrite the sentence here to something like: "Spectrometer drivers need to be installed for this service, which is explained in the python-seabreeze documentation."

docs/services/oceanoptics_spectrometer.rst
```diff
@@ -0,0 +1,42 @@
...
+Configuration
+-------------
+
+.. code-block:: YAML
+
+  spectrometer:
+    service_type: oceanoptics_spectrometer
+    simulated_service_type: oceanoptics_spectrometer_sim
+    requires_safety: false
+    interface: oceanoptics_spectrometer
```
ivalaginja: Please invert them here too.

docs/services/oceanoptics_spectrometer.rst
```diff
@@ -0,0 +1,42 @@
...
+``wavelengths``:  Wavelengths in (nm) corresponding to each pixel of the spectrometer
+
+Commands
+--------
+
+Datastreams
+-----------
+``spectra``: Spectra acquired by the camera.
+
+``is_saturating``: If the intensity as reach the maximum value of the spectrometer.
```
ivalaginja: Either "has reached", or you reformulate to: "Whether the spectrometer is saturating (``1``) or not (``0``).

ivalaginja (commented):
A couple more tiny things and typos to address, mostly because I made you write new stuff in the previous review 😬 We also definitely need to add seabreeze to the env file, see comment below.

ivalaginja (approved):
Good to go!

PR #180, opened 2024-04-16
Add ThorlabsFW102CDll service for controlling Thorlabs FW102C motorized filter wheel

a-sevin:
This service uses the Thorlabs FW102C filter wheel SDK v5.0.0.

The Thorlabs FW102C filter wheel SDK is available at https://www.thorlabs.com/software_pages/ViewSoftwarePage.cfm?Code=FW102C.

This version lets you identify the filter wheel by its serial number.

I have a few questions about this version.
1) It's only compatible with Windows. The DLL provided by Thorlabs isn't an "Assembly DLL file", so Python can't load it on Linux. Is this an issue?
2) I prefer using the standard SDK installation path as the `os.add_dll_directory` parameter instead of adding the DLL to the repository. What's the good choice ?
3) The official Thorlabs module uses the 32bit version of the DLL for Windows. Should I try to detect the OS type and load the correct one or could we admit that the 64bit is a standard? (Is anyone still using Windows 7 32-bit?)

### DIFF
--- a/catkit2/services/thorlabs_fw102c_dll/FWxC_COMMAND_LIB.py
+++ b/catkit2/services/thorlabs_fw102c_dll/FWxC_COMMAND_LIB.py
@@ -2,7 +2,7 @@
 
 # This module is a modified version of the original FWxC_COMMAND_LIB.py file from the Thorlabs FW102C filter wheel SDK v5.0.0
 
-# The Thorlabs FW102C filter wheel SDK is avalaible at https://www.thorlabs.com/software_pages/ViewSoftwarePage.cfm?Code=FW102C
+# The Thorlabs FW102C filter wheel SDK is available at https://www.thorlabs.com/software_pages/ViewSoftwarePage.cfm?Code=FW102C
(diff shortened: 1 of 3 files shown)


ivalaginja:
@a-sevin what I know about dlls in catkit2 is that we keep them off repo. Some other Thorlabs services implement an autodetection of the DLL, I think the picomotors for example (pm).

ivalaginja:
For reference, this is the implementation for the 102C in one of our outdated repositories:
https://github.com/spacetelescope/catkit/blob/c89dfbd87533b7a402d0ce8217daef5be25389c8/catkit/hardware/thorlabs/ThorlabsFW102C.py

That's also what I based the Newport filter wheel service in THD2 on. Not sure if this is helpful to you.

a-sevin:
> That's also what I based the Newport filter wheel service in THD2 on. Not sure if this is helpful to you.

The issue with this implementation is that it is not possible to know which device is connected to it. At best, the type can be identified, but not the serial number. As there are two 102C on the THD bench, this could potentially lead to difficulties at some point.

PR #182, opened 2024-04-22
Add BMC deformable mirror class

ivalaginja:
Delta-PR for the DM base class, merges into https://github.com/spacetelescope/catkit2/pull/156.

Adds a base class for the BMC that inherits from the DeformableMirrorService service. It implements the actual hardware service for BMC DMs, using the base class.

I put the voltage discretization in the base class since that is something that applies to any DM. I do this using a property so it can be accessed by either of the two classes (set in the BMC service, handled in the base service). The discretization for the surface that gets sent to the simulator is handled in the simulated service directly.


ivalaginja:
@ehpor when you have a moment, let me know what you think.

ivalaginja:
Merging to keep a single PR open for this issue.

PR #192, opened 2024-04-24
Move simple simulator and model from hicat-package2 to catkit2

### LINKED ISSUE #187: Move simple simulator and optical model into catkit2
Example code on how to use the optical model and simulator are currently contained in HiCAT-specific repository. There is a broader interest in these example scripts, so they should be moved to the catkit2 repo.

ivalaginja:
Fixes #187.

Pair PR with https://github.com/spacetelescope/hicat-package2/pull/587.

Needs testing and converting the added test into a real test that pytest understands.

### DIFF
--- a/catkit2/services/simple_simulator/simple_simulator.py
+++ b/catkit2/services/simple_simulator/simple_simulator.py
@@ -1,0 +1,58 @@
+from catkit2.simulator.simulator import Simulator
+from hicat2.simulator import SimpleOpticalModel
+
+import hcipy
+
+
+class SimpleSimulator(Simulator):
+    '''A simple simulator for the SimpleOpticalModel.
+
+    This service is not meant to be used, but rather serves as an
+    example on how to use the Simulator base class.
+    '''
+    def __init__(self):
+        super().__init__('simple_simulator')
+
+        self.alpha = 1
+
+    def open(self):
+        self.model = SimpleOpticalModel()
+        wavefronts = [hcipy.Wavefront(self.model.pupil_grid.ones() * 1e6)]
+        self.model.set_wavefronts('pre_pupil', wavefronts)
+
+        self.images = self.make_data_stream('images', 'float64', self.model.focal_grid.shape, 20)
+
+        self.update_atmosphere()
+
+    def actuate_dm(self, at_time, dm_name, new_actuators):
+        def callback():
+            self.model.dm.actuators = new_actuators
+            self.model.purge_plane('pre_turbulence')
+
+        self.add_callback(at_time, callback)
+
+    def update_atmosphere(self):
+        self.log.info('Updating atmosphere.')
+
+        t = self.time.get()[0]
+
+        self.model.atmosphere.t = t
+        self.model.purge_plane('pre_coro')
+
+        self.add_callback(t + 0.01, self.update_atmosphere)
+
+    def camera_readout(self, camera_name, power):
+        image = power.shaped
+        image = hcipy.large_poisson(image)
+        image[image > 2**16] = 2**16
+        image = image.astype('float32')
+
+        try:
+            self.testbed.science_camera.images.update_parameters('float32', image.shape, 20)
+            self.testbed.science_camera.images.submit_data(image)
+        except Exception as e:
+            self.log.error(str(e))
+
+if __name__ == '__main__':
+    service = SimpleSimulator()
+    service.run()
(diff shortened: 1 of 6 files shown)


ivalaginja:
This is ready for review, although I don't really know how to make the test make more sense, and whether it's worth it.

tests/test_optical_model.py
```diff
@@ -0,0 +1,32 @@
...
+    pupil_grid = model.pupil_grid
+
+    wf_in = hcipy.Wavefront(pupil_grid.ones())
+    model.set_wavefronts('pre_pupil', wf_in)
+    
+    test_wf = model.get_wavefronts('science_camera')[0]
+    size1 = test_wf.intensity.shaped.shape[0]
+    size2 = test_wf.intensity.shaped.shape[1]
+
+    stream = DataStream.create('images', 'mod', 'float64', [size1, size2], 20)
```
ehpor: I'm not sure why the data stream is needed here. This is a leftover from very early on, when I wanted to test an early version of the camera viewer UI.

ehpor:
> This is ready for review, although I don't really know how to make the test make more sense, and whether it's worth it.

The test looks fine to me. Basic functionality test and checks that the code doesn't crash.

catkit2/services/simple_simulator/simple_simulator.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def update_atmosphere(self):
+        self.log.info('Updating atmosphere.')
+
+        t = self.time.get()[0]
+
+        self.model.atmosphere.t = t
+        self.model.purge_plane('pre_coro')
+
+        self.add_callback(t + 0.01, self.update_atmosphere)
```
raphaelpclt, line 40: Does this recursive call work? The `update_atmosphere` puts itself in the callback?

raphaelpclt: Shouldn't be a callback function be defined here actually?

catkit2/simulator/simple_optical_model.py
```diff
@@ -0,0 +1,77 @@
...
+        outer_scale = 20
+        velocity = 0.3
+
+        Cn_squared = hcipy.Cn_squared_from_fried_parameter(r0, 1)
+        return hcipy.InfiniteAtmosphericLayer(self.pupil_grid, Cn_squared, outer_scale, velocity)
+
+    @property
+    @with_cached_result
+    def coro(self):
+        return hcipy.PerfectCoronagraph(hcipy.make_hicat_aperture(True)(self.pupil_grid))
```
raphaelpclt, line 64: Do you want to keep hicat aperture here?

tests/test_optical_model.py
```diff
@@ -0,0 +1,21 @@
...
+    wf_in = hcipy.Wavefront(pupil_grid.ones())
+    model.set_wavefronts('pre_pupil', wf_in)
+
+    for i in range(5):
+        model.atmosphere.t += 0.01
+        model.purge_plane('pre_coro')
+
+        wf_out = model.get_wavefronts('science_camera')[0]
+
+        assert np.sum(wf_out.intensity) > 0
```
raphaelpclt, line 21: I am dubious about this test - why doing 5 iterations if it's to test the same thing every time? Why not test that the `wf_out` has changed? ^^
Also I understand it's a low-key test, so I don't think that's very important.

raphaelpclt (changes requested):
Looks good overall. 
Just a few small questions, nothing serious

catkit2/simulator/simple_optical_model.py
ivalaginja: Sure, why not. It's one of the apertures from `hcipy` so there's no harm.

tests/test_optical_model.py
ivalaginja: You're totally right, but the only thing to test here is that it runs at all. The assert is rather random 🤷

catkit2/services/simple_simulator/simple_simulator.py
ivalaginja: Yes, it recursively adds a callback function that will be executed in the next iteration. In that next iteration, the execution happens and the callback creation for the following iteration happens and so on.

ivalaginja:
@raphaelpclt all fair questions, I hope I answered them sufficiently, ready for a new review. Let me know if you'd need more.

raphaelpclt (approved):
Looks good

PR #193, opened 2024-04-24
Add `miniconda-version: "latest"` to parameters of testing action

ivalaginja:
CI started failing on MacOS only, the error message it gives us says to do this to fix things.

Screenshot 2024-04-24 at 19 02 51


RemiSoummer (approved):
Also briefly discussed with @ehpor so let's merge this.  This seems to be fixing things since the tests have passed.

PR #199, opened 2024-05-15
Stop NKT emission when stopping service

### LINKED ISSUE #198: Stopping NKT service doesn't stop emission
It seems it should be a logical behavior.

raphaelpclt:
Just stopping the emission at the beginning of the ``close()`` method, in sim and on hardware. 

This PR fixes #198. 

- [x] Tested in simulation
- [x] Tested on hardware


ivalaginja (approved):
Definitely wanted behavior

PR #200, opened 2024-05-16
Add services for Thorlabs motors using Kinesis

### LINKED ISSUE #169: Add services for Thorlabs TDC001 and KDC101 cube motors

ivalaginja:
Taking over from #170. Fixes #169. See https://github.com/spacetelescope/catkit2/issues/169#issue-2205418848 for details.

Credit for writing the hardware service goes entirely to @a-sevin.

### Todo before review
- [x] Write hardware service @a-sevin 
- [x] Write service proxy @ivalaginja 
- [x] Add simulated service @ivalaginja 
- [x] Test in software @ivalaginja
- [x] Add source copyright to hardware service @ivalaginja 
- [x] Test on hardware for KDC101 motor @erinpougheon 
- [x] Test on hardware for TDC001 motor @erinpougheon 
- [x] Test on hardware with 2 motors at the same time @erinpougheon 
- [x] Update service doc page @erinpougheon


ivalaginja:
@raphaelpclt could you provide a first review on this? Some of the hardware tests are outstanding but only as additional iterations.

docs/services/thorlabs_cube_motor_kinesis.rst
```diff
@@ -0,0 +1,54 @@
+Thorlabs Cube Motors
+====================
+
+This service connects to Thorlabs TDC001 and Thorlabs KDC101 controllers in order to operate a motor.
+
```
ivalaginja, line 5: It would be good to add the link to the Kinesis software:
 https://www.thorlabs.com/software_pages/viewsoftwarepage.cfm?code=Motion_Control#

docs/services/thorlabs_cube_motor_kinesis.rst
```diff
@@ -0,0 +1,54 @@
...
+- MTS50-Z8
+- Z825B
+
+Configuration
+-------------
+
+.. code-block:: YAML
+
+    motor:
+      service_type: thorlabs_cube_motor
```
ivalaginja: Careful, the service_type changed in the meantime.

erinpougheon:
I went to the lab to perform tests. 

I run the following tests on each motor (one KDC and one TDC):
- open the service
- tests the proxy commands (move_relative and move_absolute)
- test the service command (home())
Everything worked as expected.

Then, I performed tests with the two motors linked to the PC (the TDC and the KDC).
First, I started both, and closed both. The test was successful, everything worked as expected.
Then, I started both, and moved both. The test was also successful.
After that, I started both motors, closed one, moved the other one and closed it. I did another tests the other way around (starting both, closing the second one, moving the first one and closing it). Both tests were successful too!

All the tests we wanted to do on both motors with the service are done and successful.

ivalaginja:
@ehpor @raphaelpclt if any of you have some time to review this in the coming time, the PR is ready for it.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
```diff
@@ -0,0 +1,233 @@
...
+        super().__init__('thorlabs_cube_motor_kinesis')
+
+        self.cube_model = self.config['cube_model']
+        self.stage_model = self.config['stage_model']
+        self.motor_positions = self.config['positions']
+        self.motor_type = 0
+        self.serial_number = str(self.config['serial_number']).encode("UTF-8")
+
+        # Load the Thorlabs DLLs.
+        default_dll_path = r"C:\Program Files\Thorlabs\Kinesis"
```
raphaelpclt, line 51: Wouldn't it be cleaner to use an environment variable, to set up when installing the dlls?

raphaelpclt: Wouldn't it be cleaner to use an environment variable, to set up when installing the dlls?

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
```diff
@@ -0,0 +1,233 @@
...
+            self.log.debug("Serial number: %s", self.serial_number)
+            self.log.debug("Serial number list: %s", serial_number_list.value)
+            if self.serial_number not in serial_number_list.value:
+                raise ValueError(f"Device with serial number {self.serial_number} not found in list of serial numbers: {serial_number_list.value}")
+
+        # Open the device
+        self.lib.CC_Open(self.serial_number)
+        self.lib.CC_StartPolling(self.serial_number, c_int(200))
+
+        if self.stage_model in ['MTS25-Z8', 'MTS50-Z8', 'Z825B']:
```
raphaelpclt, line 103: Any reason not to put this test earlier (in the init for example)? If the motor is not supported, it's better to fail right away, rather than starting opening stuff no?

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
```diff
@@ -0,0 +1,233 @@
...
+        self.get_current_position()
+
+    def get_current_position(self):
+        current_position = self.lib.CC_GetPosition(self.serial_number)
+        real_unit = c_double()
+        self.lib.CC_GetRealValueFromDeviceUnit(self.serial_number, current_position, byref(real_unit), 0)
+        self.log.debug("Current position: %f %s", real_unit.value, self.unit)
+        self.current_position.submit_data(np.array([real_unit.value], dtype='float64'))
+
+    def wait_for_completion(self):
```
raphaelpclt, line 216: `wait_for_completion` is only used in the homing function. Don't you want to use it in `set_current_position` too?

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
```diff
@@ -0,0 +1,109 @@
...
+            self.log.warning('Motor not moving since commanded position is outside of configured range.')
+            self.log.warning('Position limits: %f %s <= position <= %f %s.',
+                             self.min_position_config, self.unit, self.max_position_config, self.unit)
+
+    def get_current_position(self):
+        position = self.current_position.get()[0]
+        self.log.debug("Current position: %f %s", position, self.unit)
+        return position
+
+    def resolve_position(self, position):
```
raphaelpclt, line 95: `resolve_position` already exists in the proxy. Any reason to have a duplicate here?

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
```diff
@@ -0,0 +1,109 @@
...
+        The unit is given in real-life units (mm if translation, deg if rotation).
+        """
+        if self.min_position_config <= position <= self.max_position_config:
+            self.testbed.simulator.move_stage(stage_id=self.id,
+                                              old_position=self.get_current_position(),
+                                              new_position=position)
+        else:
+            self.log.warning('Motor not moving since commanded position is outside of configured range.')
+            self.log.warning('Position limits: %f %s <= position <= %f %s.',
+                             self.min_position_config, self.unit, self.max_position_config, self.unit)
```
raphaelpclt: Unless I miss something, the `self.current_position` datastream is initialized in `open`, but never updated. I guess it would have to be update in `set_current_position`

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def wait_for_completion(self):
+        # wait for completion
+        time.sleep(1)
+
+    def home(self):
+        """
+        Home the motor.
+
+        This will just sleep for a moment since there is no reason to home a simulated device.
```
raphaelpclt: Hmmm what do you mean by "homing"? I thought that would also mean "go to home position"?

raphaelpclt (changes requested):
It looks good, with a few minor questions.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: You're right! Will do that.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: Also true, will move it up.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: The motor itself behaves differently when homing and when moving position, which is why we didn't block it with this wait when it's just changing position. Let me remind myself of the details and then either justify our choice or follow your advice and put it in there as well.

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
ivalaginja: Indeed, but it serves a different purpose. Note how this `resolve_position` here only exists in the simulated service and not in the hardware service. This is because the proxy method is not available in the service itself (sim nor hardware), but we needed something to be able to read the nominal position from the service config. We only need this in the simulated service to be able to put something on the `current_position` data stream when it is created, otherwise you get an error the first time this stream is read in `move_relative` of the proxy method since the data stream is still empty. This is not needed in the hardware service since we are able to ask the device itself for its current position when we start the device and can put that on the data stream instead.

It is indeed not the cleanest way to do it, but I couldn't think of another solution to the empty data stream problem after startup of the service. This solution was adapted from the simulated XPS motors:
https://github.com/spacetelescope/catkit2/blob/6c8900abfd9f96e232a6a9749eaeee2760a5dbbe/catkit2/services/newport_xps_q8_sim/newport_xps_q8_sim.py#L25

Note how this `resolve_position` method of the service is not a service command and is not accessible from outside the service; from outside, only the proxy method can be called, so there should not be any confusion in usage.

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
ivalaginja: Huh, you're raising a good point here. This is actually not a bug - the `self.current_position` data stream gets updated in the simulator service script of our thd2 simulator, much like hicat2's (both in external repos), so we're safe, every time the simulator is called to set the simulated motor position, the data stream is updated as well.

Now, why the choice has been originally done to do it this way, I don't know. I also agree that it would be more logical to update the data stream in the simulated service's `set_current_position` as is done in the hardware service. Maybe a new issue to open and to discuss?

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
ivalaginja: Yes, it drives the motor to the zero position. In the simulated service (as on hardware), this is is a completely random position so I didn't make it actually move to zero here. Do you think I should put it in? Probably better to keep consistency...

raphaelpclt: If it's just a meaningless zero, I guess it's fine. I wasn't sure if it was re-defining the "home" position as well, or sending it to a "nominal" position. Actually, is it ever used? ^^

ivalaginja: Yeah, these motors have to be homed all the time after shut downs and similar, unfortunately. They're immediately set to their nominal positions afterwards, hence the somewhat sloppy implementation of the simulated `home()`.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: Done

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: Hm I need the variables this if block defines after the motor connection has been opened though, so I will leave this where it is. The alternative would be to make them class attributes, but not sure it's worth it. Thoughts?

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
ivalaginja: Consistency between the simulated and hardware service has always been important in `catkit2`, so I updated this to actually set the motor position to 0 and update the data stream, like the `home()` command in the hardware service does. Nice catch.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: Ok this one is tricky. If we do not wait for completion in the `set_current_position()`, it intercepts the first command and executes the second command instead, and correctly. So I am fine with this behavior and don't want to add the wait_for_completion there.

However, the method `get_current_position()` gets called too early and puts an intermediate position on the current_position data stream, which the device reads while it is moving. Which is obviously not at all what we want.

We then tried adding the `wait_for_completion()` anyway, to fix the wrong values on the data stream, but in this case it just finishes running the first command. Data stream value is still wrong. *And* it puts command 2 in a cache apparently and runs it before the next command that is sent to the motor... also not the behavior we want, so I still vote against adding the `wait_for_completion()` in `set_current_position()`.

Since this was so weird, we checked the data stream values after calling `home()` a couple of times and this is always fine, so it's something about the command data stream monitoring and getting the real device position before the device stops moving.

I don't actually understand what `wait_for_completion()` does. @a-sevin, maybe you can help out with this conundrum? What does it do and what are the values for `message_id` and `message_type` it checks against?

The minimum requirement for this fix is that the correct (final) value is put on the current_position data stream after the movement concludes.

ivalaginja: Maybe I should just regularly update the current position, like the Newport XPS service?
https://github.com/spacetelescope/catkit2/blob/6c8900abfd9f96e232a6a9749eaeee2760a5dbbe/catkit2/services/newport_xps_q8/newport_xps_q8.py#L191

@raphaelpclt what do you think about this?

a-sevin: I'm reading the tutorial labview which is not so bad. I need to check what to use it in Python:
https://www.thorlabs.com/Software/Motion%20Control/Kinesis/Kinesis-labview.pdf

This `wait_for_completion()` could be very simple: wait the `get_current_position()` stable...

ivalaginja:
I now spun off the command monitoring on a thread and update the current position in regular updates in the main(). Works fine in simulation although I didn't check the data streams, which we'll do on hardware - @erinpougheon this can be retested on hardware, where we have to make sure that both `move_absolute()` and `move_relative()` still work ok, and then do the current position data stream tests like last time (checking that they carry the correct numbers, after homing and moving (and also in general)).

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
raphaelpclt: That's a minor comment, I guess it's fine that way.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
raphaelpclt: For the `get_current_position`, yes I think the solution with the newport xps is what you want. Then I am not sure of what you want for the moving behavior. The fact that the moving function returns before completion sounds a bit weird to me, but maybe not dramatic.

erinpougheon:
I just went to the lab to test the changes on hardware. 
When I sent a home() command to the motor, I can't send another command while the motor is not stopped. When I asked for the current_position, it gives 0 as expected. 
I sent a command move_absolute(15) and then sent (before the motor stopped moving) a move_absolute(4) command. The motor stopped the first movement, and went to 4. Then, I asked for current_position and it gives what we expect (4). I did the test various times, with several values and it worked as expected everytime. When I home() the motor again, I have a current_position 0. I tried with the move_relative() command too and the current_position gives what we want. 
All tests were successful.

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
ivalaginja: Considering we now get correct values on the current_position data stream within the time scales we need (1 s), I will open a follow-up issue to look into a more rigorous handling of the data stream update inside of `set_current_position()`.

Issue: #206

ivalaginja:
@raphaelpclt we are getting the expected behavior now so this is ready for a new review. I opened a follow-up issue for a closer look into the data stream synchronization with the motor move (#206).

catkit2/services/thorlabs_cube_motor_kinesis_sim/thorlabs_cube_motor_kinesis_sim.py
raphaelpclt: Ok!

raphaelpclt (approved):
Looks good!

PR #204, opened 2024-05-22
Add performance tracing

### LINKED ISSUE #19: Performance tracing
Currently, tracing is totally unimplemented, aside from allocating port numbers in the Testbed and the tracing proxy itself.

The tracing proxy outputs the [Trace Event Format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit), originally developed by Google for Chrome. Chrome itself can do visualization of the traces with [chrome://tracing/](chrome://tracing/) or with a more refined UI Perfetto https://ui.perfetto.dev/. Example output below (for a Chrome trace).

image

These traces can be useful for analyzing latency and quickly seeing where performance is lost.

We can do tracing of:
* DataStream submitting data: a duration event from `RequestNewFrame()` to `SubmitFrame()`.
* Getting and setting properties: a duration event from start to finish.
* Executing commands: a duration event from start to finish.
* Manual traces with context manager for duration event. This may be helpful identifying computation time for wavefront controllers. The C++ variant would be RAII-based.

What is clear is that the tracing should not be "on" all the time. This would generate too much output. Likely we want some button in the GUI that can turn tracing output on and off, or a widget that includes the output file.

Also, to integrate the TracingProxy into DataStreams requires some carefulness to not impact performance. As the TracingProxy operates on ZMQ sockets, it is gonna be much slower than DataStreams.

ehpor:
image

(https://ui.perfetto.dev/ with [trace_1716350249429102200.json](https://github.com/spacetelescope/catkit2/files/15398028/trace_1716350249429102200.json) )

On simulation.

- [x] Do one test on hardware.
- [x] Test if Ctrl+C still works, since for some reason it doesn't on my Windows desktop (but develop does the same thing, so it might be something on my installation).

Fixes https://github.com/spacetelescope/catkit2/issues/19


ehpor:
First test on hardware:
image

catkit2/testbed/tracing.py
```diff
@@ -0,0 +1,218 @@
...
+                # Convert event to JSON format.
+                event_type = proto_event.WhichOneof('event')
+                if event_type == 'interval':
+                    raw_data = proto_event.interval
+                    data = {
+                        'name': raw_data.name,
+                        'cat': raw_data.category,
+                        'ph': 'X',
+                        'ts': (raw_data.timestamp - t_0) / 1000,
+                        'dur': raw_data.duration / 1000,
```
ivalaginja, line 118: Should there be a note somewhere about the units of the time stamp and durations here or is this obvious somewhere else? Same for all other instances in here that have a hard-coded unit conversion.

ehpor: Will do. There are hardcoded units everywhere, and they are fixed by the Chrome tracing format.

ehpor: Done.

ivalaginja (approved):

PR #207, opened 2024-05-31
Reset properties while acquisition in Allied Vision camera service

### LINKED ISSUE #183: Stop acquisition when required to reset certain properties in Allied Vision camera service
Some service properties can only be reset if the acquisition is temporarily stopped, for example the width and height of the ROI. This has already been implemented for the [FLIR camera service](https://github.com/spacetelescope/catkit2/blob/develop/catkit2/services/flir_camera/flir_camera.py) and we need the same for the Allied Vision camera service.

erinpougheon:
Fix #183.

Changing the Allied Vision Camera service so we will be able to change certain properties without having to stop the acquisition.

### DIFF
--- a/catkit2/services/allied_vision_camera/allied_vision_camera.py
+++ b/catkit2/services/allied_vision_camera/allied_vision_camera.py
@@ -8,6 +8,7 @@ It provides a simple interface to control the camera and acquire images.
 from __future__ import annotations
 import contextlib
 import threading
+import time
 
 import numpy as np
 
@@ -20,6 +21,32 @@ from vimba import (AllocationMode,
 )
 
 from catkit2.testbed.service import Service
+
+
+def _create_property(property_name, read_only=False, stopped_acquisition=True):
+    def getter(self):
+        with self.mutex:
+            return getattr(self.cam, property_name).get()
+
+    if read_only:
+        setter = None
+    else:
+        def setter(self, value):
+            was_running = self.is_acquiring.get()[0] > 0
+
+            if was_running and stopped_acquisition:
+                self.end_acquisition()
+
+                while self.is_acquiring.get()[0]:
+                    time.sleep(0.001)
+
+            with self.mutex:
+                getattr(self.cam, property_name).set(value)
+
+            if was_running and stopped_acquisition:
+                self.start_acquisition()
+
+    return property(getter, setter)
 
 
 class FrameHandler:
@@ -160,6 +187,9 @@ class AlliedVisionCamera(Service):
 
         self.should_be_acquiring = threading.Event()
         self.should_be_acquiring.set()
+
+        # Create lock for camera access
+        self.mutex = threading.Lock()
 
     def open(self):
         '''
@@ -323,6 +353,18 @@ class AlliedVisionCamera(Service):
         '''
         self.should_be_acquiring.clear()
 
+    exposure_time = _create_property('ExposureTime', stopped_acquisition=False)
+    gain = _create_property('Gain', stopped_acquisition=False)
+    brightness = _create_property('Brightness', stopped_acquisition=False)
+
+    width = _create_property('Width')
+    height = _create_property('Height')
+    offset_x = _create_property('OffsetX')
+    offset_y = _create_property('OffsetY')
+
+    sensor_width = _create_property('SensorWidth', read_only=True)
+    sensor_height = _create_property('SensorHeight', read_only=True)
+
     def get_temperature(self):
         '''
         Get the temperature of the camera.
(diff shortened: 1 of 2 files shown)


erinpougheon:
First of all, I had a "problem" while trying to connect the camera to the PC to run tests. Python code wasn't able to detect the camera, even though it was connected well to the PC. 
We discovered that it was due to the driver. We needed to go through NI MAX to change the Allied Vision Camera driver so we could control it though Python. 
Here is a picture of NI MAX to remember how to change the driver : 
AV_cam_driver

I added a sentence in the Allied Vision Camera service documentation saying that the camera needs the proper driver to be able to run under the service.

erinpougheon:
I ran some tests on the Allied Vision Camera in the corono room to determine which properties need the acquisition to stop to be changed : 
- width, height, offset_x, offset_y : needed to have the acquisition stopped to be changed
- exposure_time, gain : could be changed on the fly
- brightness : problem, the service told me that it has no such attribute

So, with inspiration from the FLIR camera, I changed the way properties are created, so they can all be changed without having to stop mannualy the acquisition. 
I ran some tests to verity, and they were successful, I was able to change width, height, offset_x, offset_y (and gain, exposure_time) without having to stop the service. 
However, I still have the same probelm with the brightness, the service didn't recognize it. I think that "brightness" may have another name that could be recognized, but I need to figure out which one.

catkit2/services/allied_vision_camera/allied_vision_camera.py
```diff
@@ -231,6 +248,19 @@ def make_property_helper(name, read_only=False):
...
                 self.make_property(name, lambda: getattr(self, name), lambda val: setattr(self, name, val))
 
+        # Set device values from config file (set width and height before offsets)
+        offset_x = self.config.get('offset_x', 0)
+        offset_y = self.config.get('offset_y', 0)
+
+        self.width = self.config.get('width', self.sensor_width - offset_x)
+        self.height = self.config.get('height', self.sensor_height - offset_y)
+        self.offset_x = offset_x
+        self.offset_y = offset_y
```
ivalaginja, line 258: Could you read from the configfile right here, like is done in the FLIR service? It would still set the offsets after the width and height, but we could delete the two lines in 252 and 253.

```suggestion
        self.offset_x = self.config.get('offset_x', 0)
        self.offset_y = self.config.get('offset_y', 0)
```

docs/services/allied_vision_camera.rst
```diff
@@ -3,6 +3,8 @@ Allied Vision Camera
 
 This service controls an Allied Vision camera. It is a wrapper around the Vimba SDK, which requires its installation.
 The service uses the Python API for Vimba SDK, called ``VimbaPython``.
+In order to be able to run the service on an Allied Vision Camera, the right Driver needs to be set up on the device.
+The Allied Vision Camera's driver can be changed through NI MAX.
```
ivalaginja: Let's be more specific. You can simplify the first sentence to:
"To be able to run this service, the camera needs to be set up with the right Allied Vision USB driver."
And then we would need more information included in the second sentence:
"If the device is also controlled through other interfaces, for example NI MAX, you need to switch back to the right driver through that interface."

Don't hesitate to change some of my wording if you have other ideas.

ivalaginja (commented):
Looks good to me, just requesting two minor adaptations.

Also, please don't forget to create a new issue on catkit2 stating that the brightness property of the Allied Vision does not work, and link to this PR.

catkit2/services/allied_vision_camera/allied_vision_camera.py
erinpougheon: I'm not changing this because of the fact the assignement of width and height needs the offsets in what was coded previously.

docs/services/allied_vision_camera.rst
erinpougheon: I changed the documentation according to your requests !

ivalaginja:
Looks good to me! @erinpougheon please remind me, did you test this on hardware yet? I think so, just wanted to confirm.

Btw the failing test is not due to your changes. We might have to get someone like @raphaelpclt in the loop to figure out what is happening there, I am not entirely sure I understand it.

erinpougheon:
> Looks good to me! @erinpougheon please remind me, did you test this on hardware yet? I think so, just wanted to confirm.
> 
> Btw the failing test is not due to your changes. We might have to get someone like @raphaelpclt in the loop to figure out what is happening there, I am not entirely sure I understand it.

Yes, I ran some tests on an Allied Vision Camera in the corono room last week. Since the only changes are in the documentation, the tests are still valid I think !

ivalaginja:
Great, then I will go ahead and approve.

I'll ping the RMOL folks when they start waking up to ask about the failing CI test.

ivalaginja (approved):

raphaelpclt:
Just re-ran the test that failed, and... it worked :) Don't know what happened, but you're good to go @erinpougheon

ivalaginja:
Thanks a lot @raphaelpclt!

PR #209, opened 2024-05-31
Fix cube motor service doc file name in index.rst

ivalaginja:
We forgot to update it when switching from PR https://github.com/spacetelescope/catkit2/pull/170 to PR https://github.com/spacetelescope/catkit2/pull/200.


raphaelpclt (approved):
Ok!

PR #211, opened 2024-06-04
 Reset properties while acquisition in Hamamatsu camera service

### LINKED ISSUE #194: Stop acquisition when required to reset certain properties in Hamamatsu camera service
Some service properties can only be reset if the acquisition is temporarily stopped, for example the width and height of the ROI. This has already been implemented for the [FLIR camera service](https://github.com/spacetelescope/catkit2/blob/develop/catkit2/services/flir_camera/flir_camera.py) and we need the same for the Hamamatsu camera service.

erinpougheon:
Fixes #194.

Changing the Hamamatsu Camera service so we will be able to change certain properties without having to stop the acquisition.

### DIFF
--- a/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
+++ b/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
@@ -27,9 +27,9 @@ def _create_property(property_name, read_only=False, stopped_acquisition=True):
     def getter(self):
         with self.mutex:
             if property_name != 'EXPOSURETIME':
-                return getattr(self.cam, property_name).prop_getvalue(dcam.DCAM_IDPROP.property_name)
+                return self.cam.prop_getvalue(getattr(dcam.DCAM_IDPROP, property_name))
             else:
-                return getattr(self.cam, property_name).prop_getvalue(dcam.DCAM_IDPROP.property_name) * 1e6
+                return self.cam.prop_getvalue(getattr(dcam.DCAM_IDPROP, property_name)) * 1e6
 
     if read_only:
         setter = None
@@ -45,9 +45,9 @@ def _create_property(property_name, read_only=False, stopped_acquisition=True):
 
             with self.mutex:
                 if property_name != 'EXPOSURETIME':
-                    getattr(self.cam, property_name).prop_setvalue(dcam.DCAM_IDPROP.property_name, value)
+                    self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, property_name), value)
                 else:
-                    getattr(self.cam, property_name).prop_setvalue(dcam.DCAM_IDPROP.property_name, value / 1e6)
+                    self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, property_name), value / 1e6)
 
             if was_running and stopped_acquisition:
                 self.start_acquisition()
--- a/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
+++ b/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
@@ -28,7 +28,7 @@ def _create_property(hamamatsu_property_name, read_only=False, stopped_acquisiti
         with self.mutex:
             if hamamatsu_property_name == 'EXPOSURETIME':
                 return self.cam.prop_getvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name)) * 1e6
-            elif hamamatsu_property_name in ['IMAGE_WIDTH', 'IMAGE_HEIGHT', 'SUBARRAYHSIZE', 'SUBARRAYVSIZE', 'SUBARRAYVPOS', 'SUBARRAYHPOS']:
+            elif hamamatsu_property_name in ['SUBARRAYHSIZE', 'SUBARRAYVSIZE', 'SUBARRAYVPOS', 'SUBARRAYHPOS']:
                 return int(self.cam.prop_getvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name)))
             else:
                 return self.cam.prop_getvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name))
@@ -187,6 +187,9 @@ class HamamatsuCamera(Service):
         self.log.info('Using pixel format: %s', self.current_pixel_format)
         self.cam.prop_setvalue(dcam.DCAM_IDPROP.IMAGE_PIXELTYPE, self.pixel_formats[self.current_pixel_format])
 
+        self.sensor_width = self.config.get('sensor_width', 4096)
+        self.sensor_height = self.config.get('sensor_height', 2304)
+
         # Create datastreams
         # Use the full sensor size here to always allocate enough shared memory.
         self.images = self.make_data_stream('images', 'float32', [self.sensor_height, self.sensor_width], self.NUM_FRAMES)
@@ -210,7 +213,6 @@ class HamamatsuCamera(Service):
         self.offset_x = offset_x
         self.offset_y = offset_y
 
-        self.gain = self.config.get('gain', 0)
         self.exposure_time = self.config.get('exposure_time', 1000)
         self.temperature = self.make_data_stream('temperature', 'float64', [1], 20)
 
@@ -334,8 +336,6 @@ class HamamatsuCamera(Service):
 
     gain = _create_property('CONTRASTGAIN', read_only=True)
     brightness = _create_property('SENSITIVITY', read_only=True)
-    sensor_width = _create_property('IMAGE_WIDTH', read_only=True)
-    sensor_height = _create_property('IMAGE_HEIGHT', read_only=True)
 
     def get_temperature(self):
         """
--- a/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
+++ b/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
@@ -188,7 +188,7 @@ class HamamatsuCamera(Service):
 
         # Create datastreams
         # Use the full sensor size here to always allocate enough shared memory.
-        self.images = self.make_data_stream('images', 'float32', [self.sensor_height, self.sensor_width], self.NUM_FRAMES)
+        self.images = self.make_data_stream('images', 'float32', [int(self.sensor_height), int(self.sensor_width)], self.NUM_FRAMES)
 
         self.is_acquiring = self.make_data_stream('is_acquiring', 'int8', [1], self.NUM_FRAMES)
         self.is_acquiring.submit_data(np.array([0], dtype='int8'))
@@ -258,7 +258,7 @@ class HamamatsuCamera(Service):
         # Make sure the data stream has the right size and datatype.
         has_correct_parameters = np.allclose(self.images.shape, [self.height, self.width])
         if not has_correct_parameters:
-            self.images.update_parameters('float32', [self.height, self.width], self.NUM_FRAMES)
+            self.images.update_parameters('float32', [int(self.height), int(self.width)], self.NUM_FRAMES)
 
         # Start acquisition.
         if self.cam.buf_alloc(self.NUM_FRAMES) is False:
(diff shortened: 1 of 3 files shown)


erinpougheon:
With inspiration from the FLIR camera (and Alied Vision Camera), I changed the way properties are created, so they can all be changed without having to stop mannualy the acquisition.
I will have to run some tests to verify when the Hamamatsu Camera will arrive (and also to be sure which property needs the acquisition to be stopped to be modified).
I had to make a distinction in the _create_propery method for the exposure time property, due to units that are used.
I have a doubt on how I'm using the getattr in the setter and getter (see l.29, 31, 47, 49).

A software review can be done, knowing this will require to be tested.

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -21,6 +22,38 @@
     raise
 
 
+def _create_property(property_name, read_only=False, stopped_acquisition=True):
+    def getter(self):
+        with self.mutex:
+            if property_name != 'EXPOSURETIME':
+                return getattr(self.cam, property_name).prop_getvalue(dcam.DCAM_IDPROP.property_name)
```
ivalaginja: Since `property_name` is a string, we cannot call it - this is exactly why we use `gettattr()`. And `property_name` is a property of `dcam.DCAM_IDPROP`, so that is what we have to call `getattr()` on.

The correct way to write this is, for a getter:
```python
return self.cam.prop_getvalue(getattr(dcam.DCAM_IDPROP, property_name))
```

and for a setter:
```python
self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, property_name), value)
```

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -21,6 +22,38 @@
     raise
 
 
+def _create_property(property_name, read_only=False, stopped_acquisition=True):
```
ivalaginja: I didn't point this out in the AV PR, but since this function `_create_property()` is specific to Hamamtsu cameras because of the getters and setters, lets call `property_name` -> `hamamatsu_property_name` instead. It's a bit long, but helps for understanding.

Also, could I maybe ask you to do a code-sneak and also replace that variable in the AV service with `av_property_name`?

ivalaginja (commented):
Looking good! I think I know what tripped you over with the getters and setters, so here comes a solution.

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
erinpougheon: thank you, I changed that and I understand better now !

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
erinpougheon: Done for both services ;)

ivalaginja (commented):
This looks good to me now! We will just put this PR on standby until we can test it on hardware.

ivalaginja:
@erinpougheon since the camera delivery will be delayed, could you maybe post a brief bullet-point list here what needs to be done in terms of tests once we have the new camera to wrap up this PR?

erinpougheon:
Once the camera arrives: 

• Determine which properties can be changed on the fly (go and test in an
ipython terminal, with the service open in another terminal). You need to
have to access the camera, try to change one of the properties, and then
check if it has changed. The properties to test are : exposure time, gain,
brightness, width, height, offset on x, offset on y.
• Check if the properties are well ’sorted’ in the Hamamatsu properties
branch on catkit2 (in this PR). 2 categories : the properties
that can be changed without having to stop the acquisition, or the ones
that can not.
• Test the service in the catkit2 branch, so see if we are now able to change
these properties

ivalaginja:
As an update, we received the new camera and will be integrating it in the following weeks, which means we will be able to address this PR.

ivalaginja:
We got our new camera and @johanmazoyer is working on it now.

johanmazoyer:
Works on the new camera after small corrections. 
tested all different properties:

`cam.nameprop = xxx`
and then check `cam.nameprop`

The change of property is ok for nameprop in [exposure time, width, height, offset on x, offset on y], 

The change of property ins **not** working for gain and brightness (even by restarting acquisition) but this is not suprising as these are nor parameters that can be changed for C15550-22UP. I know these are default catkit2 camera property, so they need to appear but I set them to readonly to avoid any confusion.

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -288,6 +324,19 @@ def end_acquisition(self):
         """
         self.should_be_acquiring.clear()
 
+    exposure_time = _create_property('EXPOSURETIME', stopped_acquisition=False)
+
```
ivalaginja, line 332: Too many blank lines

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -150,22 +186,9 @@ def open(self):
...
-        self.offset_y = offset_y
-
-        self.gain = self.config.get('gain', 0)
-        self.exposure_time = self.config.get('exposure_time', 1000)
-        self.temperature = self.make_data_stream('temperature', 'float64', [1], 20)
-
         # Create datastreams
         # Use the full sensor size here to always allocate enough shared memory.
-        self.images = self.make_data_stream('images', 'float32', [self.sensor_height, self.sensor_width], self.NUM_FRAMES)
+        self.images = self.make_data_stream('images', 'float32', [int(self.sensor_height), int(self.sensor_width)], self.NUM_FRAMES)
```
ivalaginja: Hmm I am not a fan of this since it duplicates code and prevents from camera class abstraction. I think it would be better to move this to where the sensor width and height properties are retrieved, i.e. into the definition of `_create_property`. There, in the getter you can add an extra case of:
```python
if hamamatsu_property_name == (IMAGE_WIDTH or IMAGE_HEIGHT):
    return int(...)
```

johanmazoyer: Yeah my thinking was that if we have an 'elif' case for each of the individual property (exposure time, IMAGE_WIDTH, etc), then there is not really a interest in creating a one for all function. But if you really think it is a good idea, i can do it your way

ivalaginja: It still gives us a layer of abstraction for the data streams that we will need for the unified camera base class to be able to deal with orientations consistently, so yes please.

johanmazoyer:
what do you think about putting gain and brightness property to -1 in the yaml file. I know these are default camera property in catkit2 but I am uneasy about having the impression you can change these parameter but it does not do anything to the camera

ivalaginja:
> what do you think about putting gain and property to -1 in the yaml file. I know these are default camera property in catkit2 but I am uneasy about having the impression you can change these parameter but it does not do anything to the camera

I would just delete the gain in the config file. This is one of the reasons to give these properties defaults in the service itself, in this way they don't have to show up in our local configuration, especially if we don't intend to change its value ever (or simply can't).

What other property are you referring to with "and property" in the yaml file?

ivalaginja:
(Also btw, please don't resolve review comments as the PR author. That functionality is supposed to help the reviewer to keep track of the individual review items.)

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
johanmazoyer: done

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -22,6 +23,41 @@
...
+                self.end_acquisition()
+
+                while self.is_acquiring.get()[0]:
+                    time.sleep(0.001)
+
+            with self.mutex:
+                if hamamatsu_property_name == 'EXPOSURETIME':
+                    self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name), value / 1e6)
+                elif hamamatsu_property_name in ['IMAGE_WIDTH', 'IMAGE_HEIGHT', 'SUBARRAYHSIZE', 'SUBARRAYVSIZE', 'SUBARRAYVPOS', 'SUBARRAYHPOS']:
+                    return int(self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name), value))
```
ivalaginja: This is a setter, not a getter, so it shouldn't return anything (as it didn't before).

docs/services/hamamatsu_camera.rst
```diff
@@ -37,27 +37,22 @@ Configuration
...
       sensor_height: 2304
       exposure_time: 8294.4
-      gain: 0
 
 Properties
 ----------
 ``exposure_time``: Exposure time of the camera in microseconds.
 
-``gain``: Gain of the camera.
+``width``: The width of the camera frames (int).
```
ivalaginja: We didn't have types here before, so why add them now? If anything, units are what should be mentioned in the docs instead. It's all pixels here so I think you can save yourself the work.

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
johanmazoyer: yes

docs/services/hamamatsu_camera.rst
johanmazoyer: done

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -22,6 +23,41 @@
...
+                self.end_acquisition()
+
+                while self.is_acquiring.get()[0]:
+                    time.sleep(0.001)
+
+            with self.mutex:
+                if hamamatsu_property_name == 'EXPOSURETIME':
+                    self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name), value / 1e6)
+                elif hamamatsu_property_name in ['IMAGE_WIDTH', 'IMAGE_HEIGHT', 'SUBARRAYHSIZE', 'SUBARRAYVSIZE', 'SUBARRAYVPOS', 'SUBARRAYHPOS']:
+                    int(self.cam.prop_setvalue(getattr(dcam.DCAM_IDPROP, hamamatsu_property_name), value))
```
ivalaginja: The PR is good to go in my opinion, but this `int()` makes no sense here. If the code works as-is then just delete it, since you're casting a function to an int. Or you're casting something (whatever is returned by `self.cam.prop_setvalue()`) to an int that you're not reusing anyway.

If you don't need to cast `value` to int before sending it to the camera (which is not being done right now), it means you could delete the entire `elif` clause here - if you have tested on hardware that all the setters actually work.

johanmazoyer: ok, let me make this correction then retest on hardware then merge

johanmazoyer:
Ok test is not totally working. 

For some reason 'IMAGE_WIDTH'/'IMAGE_EIGHT' parameters,which we always though as the width of the sensor (4096/2304), not the image width (e.g. 400/400), which has its own parameter (SUBARRAYHSIZE/SUBARRAYVSIZE). 

self.cam.prop_get_value(dcam.DCAM_IDPROP.IMAGE_WIDTH) change after I redefine SUBARRAYHSIZE, so it seems like its now equivalent to SUBARRAYHSIZE but read only. This is weird, I don't think Erin notice that. 

Therefor if we use IMAGE_WIDTH to know the size of the sensor once we initialized the camera it won't work. 

Solution found: 
IMAGE_WIDTH /IMAGE_HEIGHT now read at the begining of the initialization while they correctly indicate sensor dimensions and used to define sensor_with/sensor_height properties. After that, they are not read.

ivalaginja:
Ugh that's weird indeed, and frustrating. This is deifnitlet not expected behavior.

THD2-team:
I have now removed IMAGE_WIDTH and IMAGE_HEIGHT as property that you read from the camera and used the already written yaml param sensor_width/sensor_height that were previously unused to set the camera property of the same name.

ivalaginja (approved):
This is good to go. Thank you for working through the very annoying issue with the sensor size.

PR #212, opened 2024-06-04
Add informations about needing to create an environment variable in the Thorlabs Cubes service doc

erinpougheon:
Add informations about the envrionnement variable THORLABS_KINESIS_DLL_PATH that needs to be created in order to be able to use Thorlabs motors with Kinesis in the thorlabs_cube_motor_kinesis service.

### DIFF
--- a/docs/services/thorlabs_cube_motor_kinesis.rst
+++ b/docs/services/thorlabs_cube_motor_kinesis.rst
@@ -7,6 +7,7 @@ This service uses bits of the official vendor Python library:
 https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python
 The service also requires the installation of the Thorlabs Kinesis software:
 https://www.thorlabs.com/software_pages/viewsoftwarepage.cfm?code=Motion_Control#
+Then, to use Thorlabs cube motors with Kinesis, you need to set the THORLABS_KINESIS_DLL_PATH environment variable.
 
 Successfully tested with the following devices:
 
--- a/docs/services/thorlabs_cube_motor_kinesis.rst
+++ b/docs/services/thorlabs_cube_motor_kinesis.rst
@@ -3,7 +3,7 @@ Thorlabs Cube Motors
 
 This service connects to Thorlabs TDC001 and Thorlabs KDC101 controllers in order to operate a motor.
 
-This service uses bits of the official vendor Python library:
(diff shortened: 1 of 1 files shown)


docs/services/thorlabs_cube_motor_kinesis.rst
```diff
@@ -7,6 +7,7 @@ This service uses bits of the official vendor Python library:
 https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python
 The service also requires the installation of the Thorlabs Kinesis software:
 https://www.thorlabs.com/software_pages/viewsoftwarepage.cfm?code=Motion_Control#
+Then, to use Thorlabs cube motors with Kinesis, you need to set the THORLABS_KINESIS_DLL_PATH environment variable.
```
ivalaginja: Let's put the env variable name under `` ``, to make it stand out. In the doc, it would have to be written as:
```
``THORLABS_KINESIS_DLL_PATH``
```

docs/services/thorlabs_cube_motor_kinesis.rst
```diff
@@ -7,6 +7,7 @@ This service uses bits of the official vendor Python library:
 https://github.com/Thorlabs/Motion_Control_Examples/tree/main/Python
```
ivalaginja: Could you reformat the two links in this paragraph so that both are marked properly when rendered? There are a couple of examples of links in the Allied Vision camera service doc, you can see how to format a url there. This is especially important for the other link, on line 9, since it contains a bunch of special characters.

ivalaginja: Also, I have no clue how to do this in an rst file, but maybe you can figure out how to put a single line break after each link? I don't mean a new paragraph, but just pushing the text to the next line.
This is one of those requests that if you cannot figure out in ~15-20 minutes, please just drop it...

ivalaginja (commented):
Small change request, and some extra requests to use the chance to clean this service doc up a little.

docs/services/thorlabs_cube_motor_kinesis.rst
erinpougheon: I managed to do it !

ivalaginja (approved):
Great!

Issue #190, opened 2024-04-24
Fill in all service doc pages


ivalaginja:
Unchecked services that already have some content need to be double-checked and sometimes updated with driver information.

- [x] AimTTi PL-P - @ivalaginja - https://github.com/spacetelescope/catkit2/pull/196
- [x] AlliedVision camera - @ivalaginja - https://github.com/spacetelescope/catkit2/pull/185
- [x] BMC DM - @steigersg - https://github.com/spacetelescope/catkit2/pull/224
- [x] Camera sim - @steigersg  #252 
- [x] Empty service - @ivalaginja - https://github.com/spacetelescope/catkit2/pull/196
- [x] FLIR camera - @steigersg - https://github.com/spacetelescope/catkit2/pull/224
- [x] Newport Picomotor - @meiji274 - https://github.com/spacetelescope/catkit2/pull/197
- [x] Newport XPS - @ivalaginja - #279 
- [x] NI DAQ - @ivalaginja - https://github.com/spacetelescope/catkit2/pull/196
- [x] NKT Super K - @steigersg - https://github.com/spacetelescope/catkit2/pull/224
- [x] Omega ITHX - already existed
- [x] Safety Manual Check - @ivalaginja - #279 
- [x] Safety Monitor - already there
- [x] SNMP UPS - already there
- [x] Thorlabs CLD 101x - @ivalaginja - https://github.com/spacetelescope/catkit2/pull/196
- [x] Thorlabs filter wheel - @steigersg #252 
- [x] Thorlabs flip mount - @steigersg  #252 
- [x] Thorlabs ~picomotor~ pm -> need to rename to "power meter" - @ivalaginja - #279 
- [x] Thorlabs TSP temperature sensor - @steigersg  #252
- [x] Web Power Switch - @meiji274 - https://github.com/spacetelescope/catkit2/pull/197
- [x] ZWO camera - @meiji274 - https://github.com/spacetelescope/catkit2/pull/197

General guidelines:
- use provided template
- general description should list required drivers, Python APIs, and models a service has been tested on
- service properties can be identified with `make_property()`
- service commands can be identified with `make_command()`
- data streams can be identified with `make_data_stream()`

steigersg:
Do we need to revise the bmc_dm documentation now that we have the new DM base class?

ivalaginja:
[...]

PR #213, opened 2024-06-04
Add or fix missing service docs

ivalaginja:
I found a bunch of typos and missing services in this list, this PR fixes them.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -19,11 +19,13 @@ Catkit2
    :maxdepth: 1
    :caption: Built-In Services
 
-   service/aimtti_plp
+   services/aimtti_plp
+   services/allied_vision_camera
    services/bmc_dm
    services/camera_sim
    services/empty_service
    services/flir_camera
+   services/hamamatsu_camera
    services/newport_picomotor
    services/newport_xps_q8
    services/ni_daq
@@ -35,7 +37,7 @@ Catkit2
    services/thorlabs_cube_motor_kinesis
    services/thorlabs_fw102c
    services/thorlabs_mff101
-   services/thorlabspm
+   services/thorlabs_pm
    services/thorlabs_tsp01
    services/web_power_switch
    services/zwo_camera
--- a/docs/services/safety_manual_check.rst
+++ b/docs/services/safety_manual_check.rst
@@ -1,0 +1,19 @@
+Safety Manual Check
+===================
+
+
+Configuration
+-------------
+
+
+
+Properties
+----------
+
+
+Commands
+--------
+
+
+Datastreams
+-----------


docs/index.rst
```diff
@@ -19,11 +19,13 @@ Catkit2
    :maxdepth: 1
    :caption: Built-In Services
 
-   service/aimtti_plp
+   services/aimtti_plp
```
erinpougheon, line 22: I think the name of the service is aimtti_plp_device

docs/index.rst
```diff
@@ -19,11 +19,13 @@ Catkit2
...
+   services/allied_vision_camera
    services/bmc_dm
```
erinpougheon, line 24: I don't know if we have to add them, but every simulated service is not added, even though the camera_sim is included in the list

docs/index.rst
```diff
@@ -19,11 +19,13 @@ Catkit2
...
    services/camera_sim
```
erinpougheon, line 25: The dummy_camera service is missing here too, I don't know is this is wanted or not

erinpougheon: Later the safety_manual_check is missing too

erinpougheon (commented):
Everything seems ok!
I found some missing services yet (but I don't know if it is done on purpose or not), and the simulated services are not included too (same remark, on purpose or not?)

docs/index.rst
ivalaginja: Also true! We were initially not entirely sure whether we should include docs for the the simulated services here or not. Since a simulated service is supposed to have exactly the same behavior like a hardware service, minus all the setup that is required for a hardware service, I recently deleted all docs for simulated services in this list.

The reason the documentation for the simulated service `camera_sim` is still there because that one is somewhat special: it is a very general simulated service that can be used as the simulated service for any camera. So it seemed a good idea to give it its own doc page.

I admit these choices are somewhat arbitrary, but they are leaning on how we are currently using catkit2 and all its services. This might well be adjusted in the future

docs/index.rst
ivalaginja: You're right! However, the file that contains the service documentation is just called aimtti_plp (see [here](https://github.com/spacetelescope/catkit2/tree/develop/docs/services) for all existing files containing service docs), and this is what we want to link to from the `index.rst`.

docs/index.rst
ivalaginja: True again! I have to admit I forgot what the dummy camera service is - it looks like a stand-in service for a camera that is simpler than camera_sim. Since I don't know of any active usage anywhere, or even a reason for it, I think it's ok to not add it for now. What are your thoughts?

docs/index.rst
erinpougheon: ok perfect !

docs/index.rst
erinpougheon: Yes, I think that if it is not used at all, it's okay not including it!

ivalaginja: As for the Safety Manual Check, good catch! I didn't find it because it doesn't even have a doc page yet. I added an empty one now and added it to the list of doc pages that need to be filled in, which we're treating in issue #190.

erinpougheon (approved):

PR #221, opened 2024-06-28
Bugfix/piezo voltage limits datastream

### LINKED ISSUE #219: NI DAQ: Limit values on total voltage data stream to same limits like hardware command sent
This is where the DAC gets updated:
https://github.com/spacetelescope/catkit2/blob/98758710dc8696a6d3d3e78aa2a1bf408bf720d5/catkit2/services/ni_daq/ni_daq.py#L99

What happens is that it first uses the method `self.write_multichannel(total_voltage)` in line 106. The voltage truncation happens inside that method.
And then in line 109, it does `self.total_voltage.submit_data(total_voltage)`, which submits the non-limited voltage that was originally commanded. Which is wrong if the `write_multichannel()` actually limits the applied voltage.

The code needs to be adapted so that the voltage limitation is applied both to the command sent to hardware, and the values submitted to the `total_voltage` data stream.

erinpougheon:
Fixes #219.

### DIFF
--- a/catkit2/services/ni_daq/ni_daq.py
+++ b/catkit2/services/ni_daq/ni_daq.py
@@ -105,9 +105,6 @@ class NiDaq(Service):
         # Apply the command on the DAQ.
         self.write_multichannel(total_voltage)
 
-        # Submit these voltages to the total voltage data stream.
-        self.total_voltage.submit_data(total_voltage)
-
     def read_multichannel(self):
         '''Read voltages from all configured input channels.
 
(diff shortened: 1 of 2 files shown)


erinpougheon:
This fix works as excpected in simulation: 
image
You can see the attempt to have +5.0V on the left (server log), and the viewer, which displays directly from the total_voltage datastream, shows only 2.0 V.

erinpougheon:
Same test on hardware, and it is successful too.
image

ivalaginja (approved):

PR #222, opened 2024-07-08
Modify Allied Vision Camera service to make it work with the new VimbaX software

### LINKED ISSUE #223: Adapt Allied Vision Camera service to VimbaX
The service for AlliedVision cameras written https://github.com/spacetelescope/catkit2/pull/133 is ready uses the Vimba driver. Allied Vision recently released a new driver, called VimbaX.

Eventually, and probably not in the foreseeable future, we will want to test the service compatibility with the VimbaX driver and the respective new Python package, VmbPy: https://github.com/alliedvision/VmbPy/
(as opposed to the Python package used for Vimba, which is VimbaPython: https://github.com/alliedvision/VimbaPython)

So, the main goal of this issue is to rewrite the AlliedVisionCamera service to use VmbPy with VimbaX.

The Python API seems to be exactly the same, so it could just mean having the hardware service check which driver is installed and using the correct package accordingly. In particular:
https://github.com/alliedvision/VmbPy#differences-to-vimbapython

Full Python API manual:
https://docs.alliedvision.com/Vimba_X/Vimba_X_DeveloperGuide/pythonAPIManual.html

There's also a migration guide:
https://docs.alliedvision.com/Vimba_X/Vimba_X_DeveloperGuide/migrationGuide.html#migration-from-vimba

Python examples with Vimba X:
https://docs.alliedvision.com/Vimba_X/Vimba_X_DeveloperGuide/examplesOverview.html#vmbpy-examples

All info about VimbaX and Vimba:
https://docs.alliedvision.com/

erinpougheon:
Fixes #223 and https://github.com/ivalaginja/thd-controls/issues/108.

### DIFF
--- a/catkit2/services/allied_vision_camera/allied_vision_camera.py
+++ b/catkit2/services/allied_vision_camera/allied_vision_camera.py
@@ -17,7 +17,7 @@ from vmbpy import (AllocationMode,
                    FrameStatus,
                    PixelFormat,
                    VmbSystem,
-                   VmbSystemError
+                   VmbCameraError
 )
 
 from catkit2.testbed.service import Service
@@ -204,7 +204,7 @@ class AlliedVisionCamera(Service):
             If the camera cannot be found.
         ValueError
             If the pixel format is invalid.
-        VmbSystemError
+        VmbCameraError
             If there is an error with the camera.
         '''
         self.vmb = VmbSystem.get_instance()
@@ -219,7 +219,7 @@ class AlliedVisionCamera(Service):
 
         try:
             self.cam = self.vmb.get_camera_by_id(camera_id)
-        except VmbSystemError as e:
+        except VmbCameraError as e:
             raise RuntimeError(
                 f'Could not find camera with ID {camera_id}') from e
         self.exit_stack.enter_context(self.cam)
--- a/catkit2/services/allied_vision_camera/allied_vision_camera.py
+++ b/catkit2/services/allied_vision_camera/allied_vision_camera.py
@@ -1,7 +1,7 @@
 '''
 This module contains a service for Allied Vision cameras.
 
-This service is a wrapper around the Vimba SDK.
+This service is a wrapper around the Vimba X SDK.
 It provides a simple interface to control the camera and acquire images.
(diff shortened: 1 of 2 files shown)


erinpougheon:
I tested the VimbaX migration on the THD2 testbed.
After installing everything that is required on the thd-controller PC, I did some testes and the required changes in the service. 
I also updated the documentation file.

ivalaginja (commented):
What are the concrete tests you ran to make sure the migration works and does not introduce any issues?

erinpougheon:
> What are the concrete tests you ran to make sure the migration works and does not introduce any issues?

For now, I opened the service in a terminal, and sent some commands though an ipython terminal. I also opened the GUI and the corresponding viewer, to go and test to change the exposure time, the width of the frame, ...

ivalaginja:
> > What are the concrete tests you ran to make sure the migration works and does not introduce any issues?
> 
> For now, I opened the service in a terminal, and sent some commands though an ipython terminal. I also opened the GUI and the corresponding viewer, to go and test to change the exposure time, the width of the frame, ...

Ok great! Could you also run some of the example scripts over in thd-controls to make sure they run without any problem? I want to make sure there's no detail we didn't catch in this transition, even if it looks very non-invasive.

erinpougheon:
I ran some further tests on the Allied Vision Cameras (the visitor camera and LOWFS camera). 
1 - the ```take_exposure.py``` example. Here is a screenshot of the result, the test is successful.
image (7)
2 - the ```take_calibrated_image.py``` example. Here is a screenshot of the result, the test is successful (even if the image is saturated !)
image (8)
3 - I ran the autoexpose function, with the autoexpose_test script. The result was 252.485 us for target 0.5 and 15 percent. It is coherent and the wanted result.
4 - Finally, I ran the TT controller loop on the LOWFS camera, and it worked too :) 
image (9)

All tests are successful and the migration to VimbaX doesn't seem to cause any problem and the Allied Vision Cameras.

ivalaginja (approved):
Excellent, thanks for all the detailed test results! This is good to go.

PR #224, opened 2024-07-09
Service docs work (Steiger)

steigersg:
This is part of the team effort to fill out the documentation pages in the services module of catkit2 (see issue #190 for more information).

service docs included in this PR:
- [x] bmc_dm  
- [x] nkt_superk
- [x] FLIR camera

### DIFF
--- a/docs/services/bmc_dm.rst
+++ b/docs/services/bmc_dm.rst
@@ -45,6 +45,8 @@ Properties
 Commands
 --------
 
+None.
+
 Datastreams
 -----------
 ``total_voltage``: Map of the total voltage applied to each actuator of the DM.
--- a/docs/services/bmc_dm.rst
+++ b/docs/services/bmc_dm.rst
@@ -3,6 +3,26 @@ Boston Deformable Mirror
 
 Configuration
 -------------
+.. code-block:: YAML
+
+    boston_dm1:
+        inclination:
+          design: 0
+          calibrated:
+        clocking:
+          design: 0
+          calibrated:
+        actuator_pitch: 150.0e-6
+        num_actuators_across: 48
+        reflective_diameter: 100.0e-3
+        include_actuator_print_through: false
+        max_volts: 200
+        active_actuator_mask: !path "../data/boston/rst_dm_2Dmask.fits"
+        actuator_influence_function: !path "../data/boston/boston_actuator_for_rst48.fits"
+        actuator_print_through: !path "../data/boston/boston_mems_actuator_medres.fits"
+        flat_map_voltage: !path "../data/boston/flat_map_volts_rst48.fits"
+        meter_per_volt_map: !path "../data/boston/gain_map_rst48.fits"
+        zero_padding_factor: 2  # 0 indicates no zero-padding.
 
 Properties
 ----------
--- a/docs/services/bmc_dm.rst
+++ b/docs/services/bmc_dm.rst
@@ -40,7 +40,7 @@ Configuration
 
 Properties
 ----------
-``channels``: List of command channel names (strings).
+``channels``: List of command channel names (dict).
 
 Commands
 --------
@@ -53,5 +53,4 @@ Datastreams
 
 ``total_surface``: Map of the total amplitude of each DM actuator (meters).
 
-``channel_name``: Name of the DM channel being commanded.
-
+``channels[channel_name]``: The command per virtual channel, identified by channel name.
--- a/docs/services/bmc_dm.rst
+++ b/docs/services/bmc_dm.rst
@@ -49,8 +49,8 @@ None.
 
 Datastreams
 -----------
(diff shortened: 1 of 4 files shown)


steigersg:
I would appreciate a second opinion on the driver information to make sure that information is accurate.

docs/services/bmc_dm.rst
```diff
@@ -1,14 +1,57 @@
 Boston Deformable Mirror
 ========================
+This service operates a Boston Micromachines MEMS DM. The following Boston DMs have been tested with catkit2 thus far:
```
ivalaginja: This is actually a service for a pair of Boston DMs controlled by the same driver. Since this can be a limitation in some cases, @ehpor and I started working on a new service that is more flexible in this regard in #156. While this will not alter this service here, it might be good to explain that it is for a pair of the same type of Boston DMs, with the same number of actuators, controlled by a single driver.

I will make sure to highlight the respective changes when I write the service doc page in #156.

docs/services/bmc_dm.rst
```diff
@@ -1,14 +1,57 @@
...
+        - poke
+        - aberration
+        - atmosphere
+        - astrogrid
+        - resume
+        - flat
 
 Properties
 ----------
+``channels``: List of command channel names (strings).
```
ivalaginja: I think it's a dict?

docs/services/bmc_dm.rst
```diff
@@ -1,14 +1,57 @@
...
 
+None.
+
 Datastreams
 -----------
+``total_voltage``: Map of the total voltage applied to each actuator of the DM.
+
+``total_surface``: Map of the total amplitude of each DM actuator (meters).
+
+``channel_name``: Name of the DM channel being commanded.
```
ivalaginja: For the DAC we used ```channels[channel_name]``: The command per virtual channel, identified by channel name.` to try to make it come through that this is a dict holding several data streams - maybe you could adapt a similar solution here?

ivalaginja (commented):
This all looks right to me! I cannot helo for the FLIR drivers, but I did leave some comments on the BMC service doc.

docs/services/bmc_dm.rst
steigersg: yup!

docs/services/bmc_dm.rst
steigersg: I agree it's a good idea to make the syntax consistent - updated.

ivalaginja (approved):
Looks good to me, thanks a lot for this work! I'll leave it up to you if you want to hold off the merge until you get confirmation on the FLIR driver info or if you want to go ahead.

docs/services/bmc_dm.rst
```diff
@@ -1,14 +1,56 @@
...
 Commands
 --------
 
+None.
+
 Datastreams
 -----------
+``total_voltage``: Map of the total voltage applied to each actuator of the DM.
+
+``total_surface``: Map of the total amplitude of each DM actuator (meters).
```
ivalaginja: I just had a look while working on that other PR, and I am pretty sure that the units of the data frames on the `total_surface` data stream are in nanometers. Could you please confirm this and if it is true, update this comment?

docs/services/bmc_dm.rst
```diff
@@ -1,14 +1,56 @@
...
+
+``channels[channel_name]``: The command per virtual channel, identified by channel name.
```
ivalaginja: On the same note, we should probably specify that the commands on the data streams for the individual channels are in units of nm surface.

docs/services/bmc_dm.rst
ivalaginja: Ha, and maybe we could avoid the word "map" here? Emiel initiated a refactor in #123 to be a little stricter on the differentiation between "DM commands" and "DM maps", and even if this does not apply to this service here, it might help to stick to that in the docstrings too. Maybe something like "Array of the total amplitude..." would work? Same for the `total_voltage` data stream.

ivalaginja (changes requested):
While working on #156, I thought of a couple more clarifications that would help to have in the docs here. Sorry to throw another round of comments at you after having already approved this.

docs/services/bmc_dm.rst
steigersg: good catch! Yes, just confirmed the units are nm and so updated that and also changed "map" to "array"

ivalaginja (dismissed):
Good by me! Again back to you to tell when you're done with the FLIR docs.

steigersg:
> Good by me! Again back to you to tell when you're done with the FLIR docs.

On second thought I edited the page the FLIR docs pointed to since I found a more descriptive one. I am happy with it now though.

ivalaginja (approved):

PR #225, opened 2024-07-10
Raise a timeout error if newport picomotor is stalling

raphaelpclt:
One of the newport picomotors keeps stalling and needs to be restarted manually. By implementing this, it enables other code to catch the error and handle it properly instead of stalling the experiments. 

The reason behind the stalling has not been understood yet.

### DIFF
--- a/catkit2/testbed/proxies/newport_picomotor.py
+++ b/catkit2/testbed/proxies/newport_picomotor.py
@@ -6,7 +6,7 @@ from ..service_proxy import ServiceProxy
 
 
 MAX_TIMEOUT_FOR_CHECKING = 1000  # ms
-
+MAX_TIMEOUT_HITS = 3
 
 @ServiceProxy.register_service_interface('newport_picomotor')
 class NewportPicomotorProxy(ServiceProxy):
@@ -30,10 +30,11 @@ class NewportPicomotorProxy(ServiceProxy):
         # Get current position.
         stream = getattr(self, axis_name.lower() + '_current_position')
 
+        num_timeout_hits = 0
         # Wait until actuator reaches new position.
         waiting_start = time.time()
 
-        exception_text = 'Moving the picomotor timed out.'
+        stalling_exception_text = 'Moving the picomotor timed out.'
         if timeout is None or timeout > 0:
             while True:
                 try:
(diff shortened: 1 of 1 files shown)


ehpor:
I'm very confused by this PR. There already is an exception that gets thrown when there is a timeout (as settable in the timeout parameter, default no timeout). This PR adds a new way to raise a timeout exception, but instead of time it is based on the number of times the data stream `get_next_frame()` function raises a timeout, which is in turn based on whether frames were added to the stream, and to the timeout that is given to `get_next_frame()`, in addition to this factor of three attempts being hardcoded. This way of doing a timeout is totally indeterministic. On top of there already being a timeout exception that is handled the correct way. So I don't see the need for this PR.

ehpor:
I'm very confused by this PR. There already is an exception that gets thrown when there is a timeout (as settable in the timeout parameter, default no timeout). This PR adds a new way to raise a timeout exception, but instead of time it is based on the number of times the data stream `get_next_frame()` function raises a timeout, which is in turn based on whether frames were added to the stream, and to the timeout that is given to `get_next_frame()`. This way of doing a timeout is totally indeterministic. On top of there already being a timeout exception that is handled the correct way.

raphaelpclt:
Very good points. I don't know exactly what's going on. The timeout exception is not raised when the apodizer mount stalls. The while loop keeps going on, going into the "Datastream timeout part", where it's just a ``pass`` instruction, and then loops again. The main thing I am trying to do is to change the ``pass`` into a ``raise`` so it can be caught somewhere else. I'll take any suggestion on how to do that, but I felt it was better and more concise to catch it there than just putting a timer. 

Also, I am not sure why, but when the apodizer mount stalls, the service is stuck (the service python code goes through a ``raise``) but is still shown as "running". Maybe because it happens in a thread within the service? 🤔

ehpor:
By default there is no timeout, unless you set one. Is that what is going on? @raphaelpclt

raphaelpclt:
> By default there is no timeout, unless you set one. Is that what is going on? @raphaelpclt

Oh I got confused by the code. Makes more sense. So I guess I just need to add a default timeout?

ehpor:
> > By default there is no timeout, unless you set one. Is that what is going on? @raphaelpclt
> 
> Oh I got confused by the code. Makes more sense. So I guess I just need to add a default timeout?

Yes? Unless I'm misunderstanding something?

Or we can set the timeout locally in hicat-package2. But, setting a default timeout here of like 10 seconds would be good in any case.

raphaelpclt:
Closing this branch since another fix has been found with #232

PR #226, opened 2024-07-15
Create a simple use example for a camera

### LINKED ISSUE #189: Set up example use case and document
@erinpougheon and @pbaudoz have an interest in this, so I will try to sketch out a game plan for this.

A minimal working example would constitute a testbed server that can run a camera service. This is then extendable arbitrarily to include more services. To make this work, we'd need:
1. A testbed configuration with `testbed.yml`
2. A CLI interface that allows the user to start the testbed server, `cli_interface.py`
3. A service configuration with a single service, a camera: `services.yml`
4. The simplest GUI viewer possible. Not sure if this should be a main window + camera viewer or just  a camera viewer, but I think how we launch the GUI right now does require that.

Numbers 1-3 should probably live in a new location in the project under `catkit2/examples/single_camera`, outside of the package. Number 4 should probably be the start of adding GUI elements to `catkit2` the package, and should go into `catkit2/catkit2/user_interface`. The concrete startup and setup of this would be called in the `cli_interface.py` of the example. 

In an ideal case, a second service will be added, a simulator service which uses the already existing `catkit2/simulator/simple_optical_model.py` - or we create an even simpler one with literally just a light source and a camera.

@ehpor @raphaelpclt any thoughts on this? I would in particular like some feedback on the scope of the example and the location of the individual files.

erinpougheon:
Fixes #189.

Also tried to update the link for the Allied Vision cameras in the ```environment.yml``` file to link it to Vimba X VmbPy library.

### DIFF
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -1,0 +1,193 @@
+'''
+This module contains a service for Ocean Optics Spectrometers.
+'''
+import numpy as np
+
+from seabreeze.spectrometers import Spectrometer
+from seabreeze.spectrometers import SeaBreezeError
+
+from catkit2.testbed.service import Service
+
+
+class OceanOpticsSpectrometer(Service):
+    '''
+    Service for Ocean Optics Spectrometer.
+
+    This service is a wrapper around the python seabreeze package.
+    It provides a simple interface to control the spectrometer and acquire spectra.
+
+    Attributes
+    ----------
+    spectrometer : Spectrometer
+        The spectrometer to control.
+    serial_number : str
+        The serial number of the spectrometer.
+    model : str
+        The model of the spectrometer.
+    pixels_number : int
+        number of pixels in the spectrum.
+    interval : float
+        The interval between spectrum acquisitions in seconds.
+    spectra : DataStream
+        A data stream to submit spectra from the spectrometer.
+    is_saturating : DataStream
+        A data stream to submit whether the spectrometer is currently saturating.
+    NUM_FRAMES : int
+        The number of frames to allocate for the data streams.
+
+    Methods
+    -------
+    open()
+        Open the service.
+    main()
+        The main function of the service.
+    close()
+        Close the service.
+    take_one_spectrum()
+        Get a spectrum from the spectrometer.
+    '''
+    NUM_FRAMES = 20
+
+    def __init__(self):
+        '''
+        Create a new OceanOpticsSpectrometer service.
+        '''
+        super().__init__('oceanoptics_spectrometer')
+
+        self.serial_number = str(self.config['serial_number'])
+        self.interval = self.config.get('interval', 1)
+        self._exposure_time = self.config.get('exposure_time', 1000)
+
+        self.spectrometer = None
+        self.model = None
+        self.pixels_number = None
+
+    def open(self):
+        '''
+        Open the service.
+
+        This function is called when the service is opened.
+        It initializes the spectrometer and creates the data streams and properties.
+
+        Raises
+        ------
+        ImportError
+            If the spectrometer cannot be found.
+        '''
+
+        # Find the spectrometer
+        try:
+            self.spectrometer = Spectrometer.from_serial_number(self.serial_number)
+        except SeaBreezeError:
+            raise ImportError(f'OceanOptics: Could not find spectrometer with serial number {self.serial_number}')
+
+        # exctract attributes
+        self.model = self.spectrometer.model
+        self.pixels_number = self.spectrometer.pixels
+
+        # Define and set default exposure time.
+        self.exposure_time = self._exposure_time
+
+        # Create datastreams
+        self.spectra = self.make_data_stream('spectra', 'float32', [self.pixels_number], self.NUM_FRAMES)
+
+        self.is_saturating = self.make_data_stream('is_saturating', 'int8', [1], self.NUM_FRAMES)
+        self.is_saturating.submit_data(np.array([0], dtype='int8'))
+
+        # Create properties
+        def make_property_helper(name, read_only=False):
+            if read_only:
+                self.make_property(name, lambda: getattr(self, name))
+            else:
+                self.make_property(name, lambda: getattr(self, name), lambda val: setattr(self, name, val))
+
+        make_property_helper('exposure_time')
+        make_property_helper('wavelengths', read_only=True)
+
+    def main(self):
+        '''
+        The main function of the service.
+
+        This function is called when the service is started.
+        '''
+        while not self.should_shut_down:
+            intensities = self.take_one_spectrum()
+            self.spectra.submit_data(np.array(intensities, dtype='float32'))
+
+            if np.max(intensities) == self.spectrometer.max_intensity:
+                self.is_saturating.submit_data(np.array([1], dtype='int8'))
+            else:
+                self.is_saturating.submit_data(np.array([0], dtype='int8'))
+
+            self.sleep(self.interval)
+
+    def take_one_spectrum(self):
+        '''
+        Measure and return one spectrum.
+        '''
+        return self.spectrometer.intensities()
+
+    @property
+    def wavelengths(self):
+        '''
+        The wavelengths of the spectrometer in nm.
+
+        Returns:
+        --------
+        ndarray of float:
+            wavelengths of the spectrometer.
+        '''
+        return self.spectrometer.wavelengths()
+
+    @property
+    def exposure_time(self):
+        '''
+        The exposure time in microseconds.
+
+        Returns:
+        --------
+        int:
+            The exposure time in microseconds.
+        '''
+        return self._exposure_time
+
+    @exposure_time.setter
+    def exposure_time(self, exposure_time: int):
+        '''
+        Set the exposure time in microseconds.
+
+        This property can be used to set the exposure time of the spectrometer.
+
+        If the `exposure_time` parameter is inferior to the minimum allowed exposure time for this
+        spectrometer (set by Ocean Optics), it is set to the minimum allowed exposure time.
+        If the `exposure_time` is larger than the maximum allowed exposure time, it is set to the
+        maximum allowed exposure time.
+
+        Parameters
+        ----------
+        exposure_time : int
+            The exposure time in microseconds.
+        '''
+        int_time_range = self.spectrometer.integration_time_micros_limits
+        if exposure_time < int_time_range[0]:
+            exposure_time = int_time_range[0]
+        if exposure_time > int_time_range[1]:
+            exposure_time = int_time_range[1]
+
+        self._exposure_time = exposure_time
+        self.spectrometer.integration_time_micros(exposure_time)
+
+    def close(self):
+        '''
+        Close the service.
+
+        This function is called when the service is closed.
+        It closes the spectrometer.
+        '''
+        self.spectrometer.close()
+        self.spectrometer = None
+
+
+if __name__ == '__main__':
+    service = OceanOpticsSpectrometer()
+    service.run()
--- a/catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
+++ b/catkit2/services/oceanoptics_spectrometer_sim/oceanoptics_spectrometer_sim.py
@@ -1,0 +1,150 @@
+'''
+This module contains a service for simulation Ocean Optics Spectrometers.
+'''
+import numpy as np
+from catkit2.testbed.service import Service
+
+
+class OceanOpticsSpectrometerSim(Service):
+    '''
+    Service for simulated Ocean Optics Spectrometer.
+
+    Attributes
+    ----------
+    spectrometer : Spectrometer
+        The spectrometer to control.
+    serial_number : str
+        The serial number of the spectrometer.
+    pixels_number : int
+        Number of pixels in the spectrum.
+    interval : float
+        The interval between spectrum acquisitions in seconds.
+    spectra : DataStream
+        A data stream to submit spectra from the spectrometer.
+    is_saturating : DataStream
+        A data stream to submit whether the spectrometer is currently saturating.
+    NUM_FRAMES : int
+        The number of frames to allocate for the data streams.
+
+    Methods
+    -------
+    open()
+        Open the service.
+    main()
+        The main function of the service.
+    close()
+        Close the service.
+    take_one_spectrum()
+        Get a spectrum from the spectrometer.
+    '''
+    NUM_FRAMES = 20
+
+    def __init__(self):
+        '''
+        Create a new OceanOpticsSpectrometer service.
+        '''
+        super().__init__('oceanoptics_spectrometer_sim')
+
+        self.serial_number = str(self.config['serial_number'])
+        self.interval = self.config.get('interval', 1)
+        self._exposure_time = self.config.get('exposure_time', 1000)
+
+
+        self.pixels_number = 3000
+
+        self.spectra = None
+        self.is_saturating = None
+
+    def open(self):
+        '''
+        Open the service.
+
+        This function is called when the service is opened.
+        It initializes the spectrometer and creates the data streams and properties.
+
+        Raises
+        ------
+        RuntimeError
+            If the spectrometer cannot be found.
+        '''
+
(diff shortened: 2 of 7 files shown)


erinpougheon:
After creating a new environment as described in the README of catkit (but with a new name):
The GUI is working in simulation. There is only a camera viewer.
image


I tested the code on hardware too, on the THD2 testbed, with the visitor_camera (called science camera in the testbed_example of catkit). It worked as expected after the installation of vmbpy for the allied vision camera.
Here are two screenshots of what it looks like, when the flat is on and without the flat, with the laser on.

image (14)
image (13)

catkit2/services/simple_simulator/simple_simulator.py
```diff
@@ -20,7 +20,7 @@ def open(self):
 
         self.images = self.make_data_stream('images', 'float64', self.model.focal_grid.shape, 20)
 
-        self.update_atmosphere()
+        self.update_atmosphere
```
ivalaginja: Was this change intentional? `self.atmosphere` is a callable, so it usually should keep the parentheses in order to be called properly.

catkit2/services/simple_simulator/simple_simulator.py
```diff
@@ -51,6 +51,10 @@ def camera_readout(self, camera_name, power):
         except Exception as e:
             self.log.error(str(e))
 
+    def get_camera_power(self, camera_name):
+        wavefronts = self.model.get_wavefronts(camera_name)
+        return sum(wf.power for wf in wavefronts)
+
```
ivalaginja: You'll need a second blank line here.

testbed_example/cli_interface.py
```diff
@@ -0,0 +1,53 @@
...
+The control software for an example testbed.
+
+Usage:
+  catkit start server [--port PORT_ID] [--simulated] [--config_path PATH]...
+  catkit start gui [--port PORT_ID]
+  catkit (-h | --help)
+  catkit --version
+
+Options:
+  -p, --port PORT_ID         The port number on which the testbed server operates.
```
ivalaginja: The indentation is a little bit messed up here, suggestion:
```suggestion
  -p, --port PORT_ID       The port number on which the testbed server operates.
```

testbed_example/config/services.yml
```diff
@@ -0,0 +1,22 @@
+simulator:
+  service_type: simple_simulator
+  requires_safety: false
```
ivalaginja: Since we don't have a safety service, we might actually be able to delete this keyword. Could you remove `requires_safety` both here and in the science_camera, and check whether everything still works as expected?

erinpougheon:
For the Vimba update in the ```environment.yml```, I let the previous link, because when I try to update it, I have a GitHub error, saying that it can be install from git (which in logical regarding the documentation of vmbpy of Vimba X). My suggestion might be to delete completly this line, but I'll wait for you opinion.

testbed_example/config/services.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  offset_x: 0
+  offset_y: 0
+  width: 400
+  height: 400
+  sensor_width: 1456
+  sensor_height: 1088
+  exposure_time: 200
+  gain: 0
+  flux: 100000000
```
ivalaginja: Same for the flux, I am not sure whether the simple simulator/optical model actually uses it. Could you try running without it?

testbed_example/config/testbed.yml
```diff
@@ -0,0 +1,13 @@
...
+startup_services: []
+base_data_path:
+  default: !path "~/example_data"
+support_data_path:
+  default: !path "~/example_support_data"
+
+base_experiment_path: "{simulator_or_hardware}/{date_and_time}_{experiment_name}/"
+sub_experiment_path: "{experiment_id:03d}_{experiment_name}/"
+
+long_term_monitoring_path: "{simulator_or_hardware}/long_term_monitoring/"
```
ivalaginja: Please add a blank line at the end of this file.

testbed_example/config/testbed.yml
```diff
@@ -0,0 +1,13 @@
+default_port: 1234
+simulator:
+  service_id: simple_simulator
+startup_services: []
+base_data_path:
+  default: !path "~/example_data"
```
ivalaginja: It might be better to rename this to "catkit2_example_data".

testbed_example/config/testbed.yml
```diff
@@ -0,0 +1,13 @@
...
+support_data_path:
+  default: !path "~/example_support_data"
```
ivalaginja: And then here "catkit2_example_support_data". A bit long, but shouldn't be a problem.

testbed_example/user_interface/assets/icon.jpg
ivalaginja, line 1: I think this quick and dirty trial does not need to become any icon ^^ Can you keep the example without one for not?

ivalaginja: Worst case, you could make a blank square in Inkscape and just write "catkit2" inside of it and use that.

testbed_example/user_interface/bench_display.py
```diff
@@ -0,0 +1,222 @@
...
+            if not hasattr(self, 'open_' + info['opens']):
+                raise ValueError(f'Viewer {info["opens"]} is not known.')
+            button.clicked.connect(functools.partial(getattr(self, 'open_' + info['opens']), info['device_name']))
+
+            self.buttons.append(button)
+
+        self.make_context_menu()
+
+
+    def make_context_menu(self):
```
ivalaginja: Since this method is completely commented out, can you delete it? It might mean that you need to delete its usages, which is fine since it doesn't do anything.

ivalaginja: Or, delete most of the comments and leave just a single group of commands so that it presents an example, for example the poking action. We'd have to keep line 154 in this case too I think.

testbed_example/user_interface/bench_display.py
```diff
@@ -0,0 +1,222 @@
...
+        scaling = min(self.width() / self.schematic_size.width(), self.height() / self.schematic_size.height())
+
+        # Set positions and sizes of buttons.
+        for button in self.buttons:
+            pos = button.relative_position
+            button.move(pos[0] * self.schematic_size.width() * scaling, pos[1] * self.schematic_size.height() * scaling)
+            button.resize(scaling * 70, scaling * 70)
+
+
+
```
ivalaginja: This looks like two blank lines too many.

testbed_example/user_interface/camera_viewer.py
```diff
@@ -0,0 +1,255 @@
...
+        vbox_sidebar.addWidget(w)
+
+        # Add parameters access
+        self.p = pg.parametertree.Parameter.create(name='Camera parameters', type='group', children=params)
+        self.tree = pg.parametertree.ParameterTree(self.sidebar)
+        self.tree.setParameters(self.p, showTop=False)
+        vbox_sidebar.addWidget(self.tree)
+
+        # Set to values of camera
+        self.p.param('Exposure', 'ExposureTime').setValue(self.camera.exposure_time / 1e6)
```
ivalaginja: This is something I want to address, although maybe not in this PR. This is a hard-coded conversion factor for the units of the camera exposure time, which is not fantastic in the example. Could you open an issue that says: "Remove hard-coded exposure time scaling in simple camera viewer" and link it here?

testbed_example/user_interface/common.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @QtCore.pyqtProperty(float)
+    def handle_position(self):
+        return self._handle_position
+
+    @handle_position.setter
+    def handle_position(self, pos):
+        self._handle_position = pos
+        self.update()
+
```
ivalaginja: Let's add another blank line here.

testbed_example/user_interface/logging_display.py
```diff
@@ -0,0 +1,260 @@
...
+        self.listen_timer.setInterval(100)
+        self.listen_timer.start()
+
+    def clear_filter(self):
+        self.search_filter.setText('')
+
+        i = self.threshold_dropdown.findText('Info')
+        if i >= 0:
+            self.threshold_dropdown.setCurrentIndex(i)
+
```
ivalaginja: Another blank line.

ivalaginja (commented):
Looking good! Just some small stuff to address.

The big question we need to answer here is whether we can move some of the GUI stuff into the actual `catkit2` package. @erinpougheon how about you address the comments I left, and then we try to estimate together whether this is doable?

catkit2/services/simple_simulator/simple_simulator.py
erinpougheon: Yes it is. Before deleting the parenthesis, I had an error ```TypeError: update_atmosphere() takes 1 positional argument but 2 were given```. It is the only solution I found to solve this problem.

ivalaginja:
> For the Vimba update in the `environment.yml`, I let the previous link, because when I try to update it, I have a GitHub error, saying that it can be install from git (which in logical regarding the documentation of vmbpy of Vimba X). My suggestion might be to delete completly this line, but I'll wait for you opinion.

@erinpougheon ok. Can you open an issue for this ("Update Vimba Python package installation to VimbaX" or something like that) so that we don't forget about it?

testbed_example/config/services.yml
erinpougheon: image

After removing the safety, I have this error, so I put the lines for 'require safety' back in the config file.
The 'flux' can be removed without any problem.

testbed_example/user_interface/camera_viewer.py
erinpougheon: Here is the corresponding new issue :  https://github.com/spacetelescope/catkit2/issues/228

testbed_example/user_interface/bench_display.py
erinpougheon: I let the poking action example, and deleted the rest.

erinpougheon:
> > For the Vimba update in the `environment.yml`, I let the previous link, because when I try to update it, I have a GitHub error, saying that it can be install from git (which in logical regarding the documentation of vmbpy of Vimba X). My suggestion might be to delete completly this line, but I'll wait for you opinion.
> 
> @erinpougheon ok. Can you open an issue for this ("Update Vimba Python package installation to VimbaX" or something like that) so that we don't forget about it?

Done, here is the new issue : https://github.com/spacetelescope/catkit2/issues/229

erinpougheon:
I changed everything you requested !

testbed_example/user_interface/bench_display.py
ivalaginja: Could you add a blank line after the `# self.setContextMenuPolicy(QtCore.Qt.ActionsContextMenu)` to make it obvious that it's a general command?

catkit2/services/simple_simulator/simple_simulator.py
ivalaginja: I find that weird. Let me have a look at this and I'll come back to you, because currently I can't find an explanation for this behavior.

ivalaginja (changes requested):
One mini thing to fix, please, and the other comment is more a note to myself - consider this PR done from your side, and I will try to check why I find that one callable call weird.

ivalaginja:
While trying to figure out what is happening with the `update_atmosphere()` method, I saw that the input to adding a callback function is not only the wrong way around where you fixed it, but also in the DM method:
https://github.com/spacetelescope/catkit2/blob/5c2f7a9e2e2b665577a03b2bf1d1ed8c7e5d1316/catkit2/services/simple_simulator/simple_simulator.py#L30

Could you fix that one as well please? It is not used in the example you set up so we didn't notice, but since we're already on it, let's fix that one too.

catkit2/services/simple_simulator/simple_simulator.py
ivalaginja: @ehpor any clue what might be happening here? @erinpougheon could you post the error message you're getting for this here please?

erinpougheon: image (13)

raphaelpclt: Hmmm I think you're avoiding the error, but I don't think the atmosphere works either. My intuition is that it just doesn't do anything. 

If you look at the simulator.py code here, https://github.com/spacetelescope/catkit2/blob/a7c17df19636676151d965e67d0085367ba14769/catkit2/simulator/simulator.py#L131C13-L131C41, it calls the callback function with time as an argument. So it will call ``self.update_atmosphere(t)``, that has 2 arguments (self, and t). You would need to define ``def update_atmosphere(self, t)`` or something like that.

Also, looking at the code, I feel the ``simple_simulator`` on develop is bugged.

erinpougheon:
@ivalaginja changed the simulator and optical model used in the example, so we don't have the same problem with the ```update_atmosphere()```. (I think we can maybe open a new issue to just adress this simulator problem, I don't know).

Now, the example is working in simulation (evn though there is no light on it, because light is not implmented in the new example of simulation). Here is the result on the simulated server:
image

Hardware test on the THD2 testbed: Everything works as expected. I was able to have the viewer (we can see things because I turned on a laser). I could change the ROI, the gain, the exposure time etc.
image (14)


I think this PR is ready for final review and can be merged. I added the parenthesis back in the update_atmosphere function, in case we want it to be a separate issue.

testbed_example/example_optical_model.py
```diff
@@ -30,6 +30,10 @@ def pupil_grid(self):
 
         return hcipy.make_uniform_grid(dims, size)
 
+    @property
+    def focal_grid(self):
+        return hcipy.make_focal_grid(12, 32)
+
```
ivalaginja: Do you think wherever you needed the `focal_grid` you could just use the `detector_grid` below? I would like the grid parameters to be read from the configuration, and I was trying to call things in the detector plane (where the focal grid is) with "detector" to keep things more or less consistent.

testbed_example/services/example_simulator/example_simulator.py
```diff
@@ -12,7 +12,7 @@ def __init__(self):
         super().__init__('example_simulator')
 
     def open(self):
-        self.model = ExampleOpticalModel()
+        self.model = ExampleOpticalModel(self.testbed.config)
```
ivalaginja: Nice catch on this - I also think it needs to be `self.testbed.config['simulator']`? Otherwise I am not sure what pool of configurations it tries to read, between all the different configuration files the example has. It might be automatically choosing the right one, but it's better to be explicit.

erinpougheon: I made it go on the simulator file in the optical model, but I can defenitly change that you're right.

erinpougheon:
image

Here is the viewer (in simulation, because we changed nothing that has to do with the hardware) after the changes, everything works.

ivalaginja (dismissed):
Good by me. Please wait with merging until we hear back from @raphaelpclt.

testbed_example/cli_interface.py
```diff
@@ -0,0 +1,53 @@
...
+Usage:
+  catkit start server [--port PORT_ID] [--simulated] [--config_path PATH]...
+  catkit start gui [--port PORT_ID]
+  catkit (-h | --help)
+  catkit --version
+
+Options:
+  -p, --port PORT_ID      The port number on which the testbed server operates.
+                          Defaults to the port specified in the config file.
+  --simulated             Whether the testbed server should be run in simulated mode or not.
```
raphaelpclt: I am not sure - shouldn't ``--simulated`` be the default option or hardcoded? This example will never be running not simulated I guess.

testbed_example/config/testbed.yml
```diff
@@ -0,0 +1,15 @@
...
+startup_services: []
+base_data_path:
+  default: !path "~/catkit2_example_data"
+support_data_path:
+  default: !path "~/catkit2_example_support_data"
+
+base_experiment_path: "{simulator_or_hardware}/{date_and_time}_{experiment_name}/"
+sub_experiment_path: "{experiment_id:03d}_{experiment_name}/"
+
+long_term_monitoring_path: "{simulator_or_hardware}/long_term_monitoring/"
```
raphaelpclt, line 15: These path are not relevant in this example

raphaelpclt (commented):
The code looks good overall. A few things can be removed (like experiment paths in the config - catkit2 doesn't implement experiment structures). Still pending discussion on what to let public or not.

PR #231, opened 2024-07-18
Create a Fourier Transform Viewer with plots

### LINKED ISSUE #230: Create a Fourier Transform Viewer
Based on #226 
We want to create a new viewer, called 'FT Viewer'. Can be accessed by the context menu (right click on the layout).
On this viewer, we want to have a context menu to choose the camera on which we want to have this viewer.
On this viewer, we want to have a frame of the camera, as in the camera viewer, with the possibility to stop acquisition etc as in the camera viewer.
We also want to have a viewer of the FT module.
Below those two viewers, we want to have a curve of the pixels that can be choosen with a cross that we can move on the camera or FT viewer. Each plot will have 2 curves (on for the y axis, one for the x axis), and there will have 2 of these plots (one under the camera frame, one under the FT module frame).

erinpougheon:
Fixes #230.

### DIFF
--- a/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
+++ b/catkit2/services/oceanoptics_spectrometer/oceanoptics_spectrometer.py
@@ -1,0 +1,193 @@
+'''
+This module contains a service for Ocean Optics Spectrometers.
+'''
+import numpy as np
+
+from seabreeze.spectrometers import Spectrometer
+from seabreeze.spectrometers import SeaBreezeError
+
+from catkit2.testbed.service import Service
+
+
+class OceanOpticsSpectrometer(Service):
+    '''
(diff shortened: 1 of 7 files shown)


erinpougheon:
image (15)

The FT Viewer is now operational. There is a viewer on the top left with the camera frame etc (exactly the same thing than the 'classic' camera viewer). There are 2 lines (one along the x axis, the other one along the y axis) that can be moved or the positions can be set. On the bottom left, underneath this camera viewer, we can see two plots, that represent the pixels lines where the two red and blue lines are on the frame. These plots update when we move the line. On the right side of the viewer, a Fourier Transform module viewer (in real time with the camera), the same two lines and the plots.
I tested on simulation and it worked, and the screenshot is the hardware test that works too.

PR #232, opened 2024-07-18
Fix newport picomotor failures

raphaelpclt:
Casting to integers twice before submitting to the datastream somehow prevents the failures to happen.  

Tested on HiCAT with 1000 target acquisition experiments back to back without problem (vs. usually fails after ~10 / 15 TA experiments). 

The why has not been understood yet. 

Mainly solves the issue in https://github.com/spacetelescope/hicat-package2/issues/435, but will not close it since we do not understand what is going on exactly.


RemiSoummer:
we are going to merge this because it works, but note to @ehpor and also @a-sevin since this is very weird and might be indication of some low-level issue that might pop up elsewhere.

RemiSoummer (approved):

PR #235, opened 2024-08-14
Add a service for Thorlabs MCLS1 light source

raphaelpclt:
Mostly based on catkit1 and on NKT service. Minimal proxy to mimic interface with other light sources.


RemiSoummer (approved):
Looks good, and works well on hardware, I tested it from a notebook and GUI at the same time.  We can come back to if if we find some issues later.

PR #239, opened 2024-08-23
Fix testbed port numbers

### LINKED ISSUE #240: Testbed port numbers are hardcoded and not retrieved from the testbed
In some places of the code base, the port numbers for the log distribution system, data log distribution system and trace message distribution system are hardcoded as offsets to the testbed port numbers (ie. testbed.port + 1). These ports are available from the testbed (via the testbed proxy) and should preferable be retrieved from there.

ehpor:
Testbed port numbers were hardcoded in some places. This PR fixes that.

Fixes #240
Required for #204

### DIFF
--- a/catkit2/testbed/experiment.py
+++ b/catkit2/testbed/experiment.py
@@ -71,7 +71,8 @@ class Experiment:
                 logging.getLogger('matplotlib').setLevel(logging.WARNING)
 
                 # Set up log forwarder.
-                log_forwarder = LogForwarder('experiment', f'tcp://{self.testbed.host}:{self.testbed.logging_ingress_port}')
+                log_forwarder = LogForwarder()
+                log_forwarder.connect('experiment', f'tcp://{self.testbed.host}:{self.testbed.logging_ingress_port}')
 
                 # Set up log writer.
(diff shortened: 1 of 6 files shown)


steigersg (approved):
Looks good to me - I can't speak much to the C code since I'm less familiar with it, but nothing jumps out as concerning.

ivalaginja:
@ehpor were there any changes necessary for this on hicat-package2? When running on a catkit2 version that includes this PR, I get errors for thd-controls when running trying to start the server hardware (not the actual THD2 PC but an adjacent one) but not in simulation (on my Mac, and Axel's). I'll need to go retrieve the exact error message.

ehpor:
I don't think so. But I'll check soon. The error message would be helpful.

Maybe some of the GUI elements?

johanmazoyer:
This is the error we are getting on Windows 10, on hardware. It happens when launching server. 
Screenshot 2024-10-29 at 17 29 41

ehpor:
Did you recompile catkit2 after pulling the latest update?

ivalaginja:
Definitely not. Let me do that and try again...

ivalaginja:
Problem solved, thanks @ehpor.

PR #242, opened 2024-08-29
Bugfix/mcls1 sim

### LINKED ISSUE #238: thorlabs_mcls1_sim service is non functional
In the thorlabs_mcls1_sim service in the init it is passing `super().__init__('thorlabs_mcls1')` when it should be `super().__init__('thorlabs_mcls1_sim')`

steigersg:
Fixes #238. 

Restores functionality to the `ThorlabsMcls1Sim` service. This includes:

- Fixing mislabeled service type.
- Adding `light_source_data` property to the `Simulator` service.
- Adjusting some of the getters and setters in the `ThorlabsMcls1Sim` service itself. 

**Todo:**

- [x] Test on simulator
- [x] Make sure is consistent with refactoring happening in #243

### DIFF
--- a/catkit2/services/thorlabs_mcls1_sim/thorlabs_mcls1_sim.py
+++ b/catkit2/services/thorlabs_mcls1_sim/thorlabs_mcls1_sim.py
@@ -104,9 +104,6 @@ class ThorlabsMcls1Sim(Service):
     def get_temperature(self):
         return self.config['target_temperature']
 
-    def get_temperature(self):
-        return self.config['target_temperature']
-
 
 if __name__ == '__main__':
     service = ThorlabsMcls1Sim()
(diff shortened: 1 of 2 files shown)


catkit2/services/thorlabs_mcls1_sim/thorlabs_mcls1_sim.py
```diff
@@ -117,7 +117,13 @@ def set_target_temperature(self, value):
         pass
 
     def get_power(self):
-        return self.testbed.simulator.light_source_data[self.id + '_power']
+        try:
```
steigersg: This is to circumvent the case where this key isn't set by the time this is called - need to figure out why that is.

ehpor (approved):
LGTM and tested.

PR #244, opened 2024-08-29
Add trace intervals for several services

ehpor:
Depends on https://github.com/spacetelescope/catkit2/pull/204.

### DIFF
--- a/catkit2/services/flir_camera/flir_camera.py
+++ b/catkit2/services/flir_camera/flir_camera.py
@@ -5,6 +5,7 @@ from PySpin import PySpin
 import numpy as np
 
 from catkit2.testbed.service import Service
+from catkit2.testbed.tracing import trace_interval
 
 def _create_property(flir_property_name, read_only=False, stopped_acquisition=True):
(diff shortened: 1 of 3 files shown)


catkit2/services/flir_camera/flir_camera.py
```diff
@@ -209,17 +210,18 @@ def acquisition_loop(self):
                         break
                     raise
 
-                try:
-                    if image_result.IsIncomplete():
-                        continue
+                with trace_interval('processing frame'):
```
ivalaginja, line 213: Why did you start the context manager here and not earlier, e.g. before the outer `try` in line 199?

ehpor: Because that includes waiting for the camera frame itself, which isn't processing it. I'm starting the trace right after we retrieved the camera frame.

ivalaginja: Understood, thanks!

ivalaginja (approved):
Looking good

PR #252, opened 2024-10-17
Feature/steiger service docs round2

steigersg:
Another round of service docs for catkit2 (see issue https://github.com/spacetelescope/catkit2/issues/190 for more information).

Services included:

- Thorlabs temperarure Sensor
- Thorlabs filter wheel
- Thorlabs flip mount 
- Simulated Camera

### DIFF
--- a/docs/services/camera_sim.rst
+++ b/docs/services/camera_sim.rst
@@ -3,22 +3,24 @@ Camera Simulator
 This service operates a simulated camera. This service is meant to mimic, but does not actually
 control, a hardware camera.
 
-When applicable, all services have a corresponding simulated service to be able to test control
-software before commanding actual hardware devices.
+This service can be used to simulate any hardware camera service since they are all written consistently.
 
 Configuration
 -------------
 .. code-block:: YAML
 
-    camera:
-      service_type: zwo_camera
+    camera1:
+      service_type: my_camera
       simulated_service_type: camera_sim
-      interface: hicat_camera
+      interface: camera
       requires_safety: false
 
+      # Keys used only by hardware service.
       device_name: ZWO ASI178MM
       device_id: 4
       well_depth_percentage_target: 0.65
+
+      # Keys used by simulated and hardware service.
       exposure_time: 1000
       width: 1680
       height: 1680
--- a/docs/services/camera_sim.rst
+++ b/docs/services/camera_sim.rst
@@ -1,14 +1,59 @@
 Camera Simulator
 ================
+This service operates a simulated camera. This service is meant to mimic, but does not actually
+control, a hardware camera.
+
+When applicable, all services have a corresponding simulated service to be able to test control
+software before commanding actual hardware devices.
 
 Configuration
 -------------
+.. code-block:: YAML
+
+    camera:
+      service_type: zwo_camera
+      simulated_service_type: camera_sim
+      interface: hicat_camera
+      requires_safety: false
+
+      device_name: ZWO ASI178MM
+      device_id: 4
+      well_depth_percentage_target: 0.65
+      exposure_time: 1000
+      width: 1680
+      height: 1680
+      offset_x: 336
+      offset_y: 204
+      gain: 0
 
 Properties
 ----------
+``exposure_time``: Simulated exposure time (in microseconds) of the camera.
+
+``gain``: Simulated gain of the camera.
+
+``width``: The width of the camera frames.
+
+``height``: The height of the camera frames.
+
+``offset_x``: The x offset of the camera frames on the sensor.
+
+``offset_y``: The y offset of the camera frames on the sensor.
+
+``sensor_width``: The width of the simulated sensor.
+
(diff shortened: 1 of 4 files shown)


docs/services/camera_sim.rst
```diff
@@ -1,14 +1,59 @@
 Camera Simulator
 ================
+This service operates a simulated camera. This service is meant to mimic, but does not actually
+control, a hardware camera.
+
+When applicable, all services have a corresponding simulated service to be able to test control
+software before commanding actual hardware devices.
```
ivalaginja: I think this should go into a different part of the docs, where the services are describe, so I think you can delete it here. Instead of it, you could say that this simulated camera service can be used as the simulated service for any of the hardware camera services, since they're all written consistently.

docs/services/camera_sim.rst
```diff
@@ -1,14 +1,59 @@
...
 
 Configuration
 -------------
+.. code-block:: YAML
+
+    camera:
+      service_type: zwo_camera
+      simulated_service_type: camera_sim
+      interface: hicat_camera
```
ivalaginja: I think better just `camera` here since that one actually lives inside of catkit2 itself.

docs/services/camera_sim.rst
```diff
@@ -1,14 +1,59 @@
...
+      requires_safety: false
+
+      device_name: ZWO ASI178MM
+      device_id: 4
+      well_depth_percentage_target: 0.65
```
ivalaginja, line 21: Could you put in a blank space after this line, so indicate the difference between the keys required by the camera_sim service (exposure_time, gain, etc.) and the keys that are used by the hardware service?

docs/services/thorlabs_fw102c.rst
```diff
@@ -1,14 +1,38 @@
...
 
 Commands
 --------
+None.
 
 Datastreams
 -----------
+``position``: The commanded position of the filter wheel. This can have value 1, 2, 3, 4, 5, 6 indicating those respective positions.
+
+``current_position``: The current position of the filter wheel. This can have value 1, 2, 3, 4, 5, 6 indicating those respective positions.
```
ivalaginja: I think the positions are dynamic and depend on whatever is put in the service configuration.

docs/services/thorlabs_fw102c.rst
```diff
@@ -1,14 +1,38 @@
...
+.. code-block:: YAML
+
+    filter_wheel:
+        service_type: thorlabs_fw102c
+        simulated_service_type: thorlabs_fw102c_sim
+        requires_safety: false
+
+        visa_id: ASRL6::INSTR
+
+        filters:
```
ivalaginja: Are these positions resolved by their name in the fw proxy as well? If so, might be worth adding the same comment in here (`# named positions resolved by the proxy.`) like in the example service config of the flip mount.

docs/services/thorlabs_mff101.rst
```diff
@@ -29,4 +30,6 @@ Commands
 
 Datastreams
 -----------
-``position``: the current (commanded) position of the flip mount. This can be set to either 1 or 2, indication position 1 or 2. Other values will be silently ignored.
+``current_position``: the current position of the flip mount. This can be set to either 1 or 2, indicating position 1 or 2. Other values will be silently ignored.
```
ivalaginja:
```suggestion
``current_position``: The current position of the flip mount. This can be set to either 1 or 2, indicating position 1 or 2. Other values will be silently ignored.
```

docs/services/thorlabs_mff101.rst
```diff
@@ -29,4 +30,6 @@ Commands
...
+
+``commanded_position``: the commanded position of the flip mount. This can be set to either 1 or 2, indicating position 1 or 2. Other values will be silently ignored.
```
ivalaginja:
```suggestion
``commanded_position``: The commanded position of the flip mount. This can be set to either 1 or 2, indicating position 1 or 2. Other values will be silently ignored.
```

ivalaginja (commented):
Thanks for doing this! Just some minor comments to hopefully clarify things further.

docs/services/camera_sim.rst
steigersg: good call, updated.

docs/services/camera_sim.rst
steigersg: done, also added comments to better make that distinction.

docs/services/thorlabs_fw102c.rst
steigersg: good idea! added

docs/services/thorlabs_fw102c.rst
steigersg: you're right, I was assuming the maximum number of positions for the fw but it should depend on the config, updated.

ivalaginja (approved):
Good stuff! Thanks so much again!

PR #255, opened 2024-10-18
Feature/4d service

### LINKED ISSUE #254: Add service for 4D AccuFiz interferometer
Catkit1 had a 4d accufiz interferometer hardware service and there is need for the service for future experiments using catkit2.

lanemeier7:
closes #254 and creates an accufiz_interferometer service for catkit2

Requires running 4d Web Services running and 4Sight Focus Software in listening mode when running actual hardware.

example service.yml entry:
``` .yml
accufiz_interferometer:
  service_type: accufiz_interferometer
  simulated_service_type: accufiz_interferometer_sim
  interface: camera
  requires_safety: false
  #sim_data: C:/path/to/example.h5
  mask: C:/path/to/4d.mask
  server_path: C:/path/to/data
  local_path: C:/path/to/data
  ip_address: localhost:8080
  num_avg: 2

```

Example testbed script:
``` .py
import pathlib
import os
import matplotlib.pyplot as plt
import numpy as np
from catkit2 import TestbedProxy, read_config_files

config_path = pathlib.Path(os.path.join('tests/', 'config'))
config_files = config_path.resolve().glob('*.yml')
config = read_config_files(config_files)

port = config['testbed']['default_port']

testbed = TestbedProxy('127.0.0.1', port)
cam = testbed.accufiz_interferometer
img = np.mean(list(cam.take_raw_exposures(num_exposures=2)), axis=0)

plt.imshow(img, cmap='inferno')
plt.colorbar()
plt.show()
```


catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,151 @@
...
+import numpy as np
+from scipy import ndimage
+import math
+from astropy.io import fits
+from glob import glob
+from catkit2.testbed.service import Service
+from catkit2.testbed.service_proxy import ServiceProxy
+from catkit2.utils import rotate_and_flip_image
+
+@ServiceProxy.register_service_interface('accufiz_interferometer')
```
raphaelpclt: This decorator is to register a service proxy, but used on a Service here. You'll have to create a service proxy in `catkit2/testbed/proxies`. Here is an example:
https://github.com/spacetelescope/catkit2/blob/f9d89aad0ef634a6290e920471c6cb7c0a1e55db/catkit2/testbed/proxies/camera.py#L6

lanemeier7: Would the cleanest course of action be for me to make this a ServiceProxy class and move this to from the services to the proxies folder? Is there any advantage of having a service over a serviceproxy?

raphaelpclt: This is the service, and it's fine as it is (I haven't looked in details yet, but looks good overall). On top of that, you'll need to make a proxy. The service runs on its own process in a real time fashion. To prevent overloads that do not require real time, we use proxies that run on the client side, that are basically the interface with the service. The proxy stores the config so it doesn't have to be communicated through tcp at every call for example, or stores some convenience functions that do more complex computation.

lanemeier7: Ok - I have made the accufiz_interferometer inherit the Service class and it also implements the Camera ServiceProxy. See updated test and service.yml examples.

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,196 @@
...
+            raise RuntimeError(f"{self.config_id}: Failed to take data - {resp.text}.")
+
+        filename = str(uuid.uuid4())
+        server_file_path = os.path.join(self.server_path, filename)
+        local_file_path = os.path.join(self.local_path, filename)
+
+        #  This line is here because when sent through webservice slashes tend
+        #  to disappear. If we sent in parameter a path with only one slash,
+        #  they disappear
+        server_file_path = server_file_path.replace('\\', '/')
```
raphaelpclt, line 197: I remember we had os-dependant issues with that. What OS did you try it on?

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,196 @@
...
+        local_file_path = local_file_path if local_file_path.endswith(".h5") else f"{local_file_path}.h5"
+
+        self.log.info(f"{self.config_id}: Succeeded to save measurement data to '{local_file_path}'")
+
+        mask = np.array(h5py.File(local_file_path, 'r').get('measurement0').get('Detectormask', 1))
+        img = np.array(h5py.File(local_file_path, 'r').get('measurement0').get('genraw').get('data')) * mask
+
+        self.detector_masks.submit_data(mask.astype(np.uint8))
+
+        image = self.convert_h5_to_fits(local_file_path, rotate=0, fliplr=True, mask=mask, img=img, create_fits=self.save_fits)
```
raphaelpclt: Would it make sense to read the `rotate` and `fliplr` from the config? Idk how setup dependent these are.

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,196 @@
...
+
+                if not has_correct_parameters:
+                    self.images.update_parameters('uint16', img.shape, 20)
+                self.log.info('requesting new frame')
+                frame = self.images.request_new_frame()
+
+                frame.data[:] = img
+                self.log.info(f'frame {frame.id} acquired')
+                self.images.submit_frame(frame.id)
+                time.sleep(0.05)
```
raphaelpclt: That procedure to submit into a datastream is correct. But that's very low level. Unless there's a specific reason to use that, you can use the more convenient `self.images.submit_data()` function. That will create a new frame and submit it automatically.

catkit2/utils.py
```diff
@@ -0,0 +1,16 @@
...
+    :param theta: Rotation in degrees of the mounted camera, only these discrete values {0, 90, 180, 270}
+    :param flip: Boolean for whether to flip the data using np.fliplr.
+    :return: Converted numpy array.
+    """
+    data_corr = np.rot90(data, int(theta / 90))
+
+    if flip:
+        data_corr = np.fliplr(data_corr)
+
+    return data_corr
```
raphaelpclt: Hum - I think this might overlap at some point with #124. The general idea is that camera flips should be handled at a very low level, and then not touched anymore. And this should be taken care of by a coming camera base class. So even if this function is very relevant for you, I don't think it should be part of a `utils` file in the catkit2 context. 
TLDR; I'd prefer this to be somewhere else - e.g. as static method in the service? @ehpor @ivalaginja @steigersg any opinion?

catkit2/utils.py
```diff
@@ -0,0 +1,16 @@
+import numpy as np
+
+def rotate_and_flip_image(data, theta, flip):
+    """
+    Converts an image based on rotation and flip parameters.
+    :param data: Numpy array of image data.
+    :param theta: Rotation in degrees of the mounted camera, only these discrete values {0, 90, 180, 270}
+    :param flip: Boolean for whether to flip the data using np.fliplr.
+    :return: Converted numpy array.
```
raphaelpclt: We use the numpy docstring format (some non-compliant docstrings might exist in the repos, but we try to get rid of them) - could you change to that format?
https://numpydoc.readthedocs.io/en/latest/format.html

raphaelpclt (changes requested):
Very nice! I have a few comments still, on datastream submission, on the relevance of the `utils.py` file and a doctstring format. 

One more general questions still: no `close` method have been implemented in the service. I have no experience with the 4d at all, but just checking. There's no "sleep mode" or so we would want to put it in while closing the service?

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
lanemeier7: running Windows. Mostly to make it look pretty in the log as path.join sometimes mixes \\ with / on Windows.

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
lanemeier7: Sounds good. I can make this change, I was copying the dummy camera example. https://github.com/spacetelescope/catkit2/blob/87c27be0f63707c360c85e64768d715c61521120/catkit2/services/dummy_camera/dummy_camera.py#L89-L107

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
lanemeier7: Yes, can make them configurable

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
raphaelpclt: Oh yes, the dummy camera is probably one of the early piece of code written by Emiel. New things have been added since then 😅

catkit2/utils.py
lanemeier7: Can do!

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
raphaelpclt: The `zwo_camera.py` service is probably more up to date.

catkit2/utils.py
lanemeier7: I have no problem making it a static method. I saw this was a util in catkit1 so I copied this policy.
My 2 cents is that having static methods inside services makes for a lot of copy pasted code and violation of the don't repeat yourself rule. 

On a slightly separate but similar topic, it would be nice if the simulated service modules could inherit the non -simulated hardware service classes and just override the stuff that needs to be different (like fake communication or fake data generation). I quickly tried this and things were unhappy in my first attempt.

ehpor: @lanemeier7 Catkit1 did exactly this. We consciously chose not to do this for catkit2, for two reasons:
* The lever of overriding is an important consideration. Catkit1 mocked the actual API coming from the hardware supplier. This is the lowest level of mocking that we can reasonably do. However, some of the low-level APIs are quite extensive and require a large amount of work to fully override. Catkit2 however mocks at the service level. This allows using one simulated service for multiple hardware services, if they are similar enough (see for example the camera, where `camera_sim` is the simulated service for now three hardware camera services.
* We want to strictly avoid loading of hardware libraries during imports. This allows for running of simulations on machines that do not have hardware APIs installed.

We can still do what you are asking for, in a separate way. Take a look at how the `bmc_deformable_mirror_hardware` and `*_sim` services are implemented. They use a common service base that implements common code between simulated software and hardware.

ehpor: As for this function itself: I personally don't see a reason to have a function for effectively two lines of code, each executing a single Numpy function. Many people will be aware of the Numpy function, but not of the util function. So those people would likely use the two numpy functions, rather than use this util function. So, unless I was using this function >3-5 times inside my service, I'd likely not have it be a function at all, if I was writing this.

That said, I totally concede that _this is largely subjective and my personal preference, so I'd be fine with anything_.

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,270 @@
...
+    Rotate and/or flip the image data.
+
+    Args:
+        data (numpy.ndarray): Numpy array of image data.
+        theta (int): Rotation in degrees.
+        flip (bool): If True, flip the image horizontally.
+
+    Returns:
+        numpy.ndarray: Modified image after rotation and/or flip.
+    """
```
raphaelpclt: This not exactly numpy docsrings. This would have to be stg like:

```
Parameters
----------
data : numpy.ndarray
    Numpy array of image data
theta : int
    Rotation in degrees.
flip : bool
    If True, flip the image horizontally

Returns
-------
numpy.ndarray
    Modified image after rotation and/or flip.
```

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        HTTP GET request.
+
+        Args:
+            url (str): URL to send the GET request to.
+            params (dict, optional): Parameters for the request. Defaults to None.
+
+        Returns:
+            resp: response object.
+        """
```
raphaelpclt: Same comment for the docstring, not exactly numpy format

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,270 @@
...
+        HTTP POST request.
+
+        Args:
+            url (str): URL to send the POST request to.
+            data (dict, optional): Data to send in the request. Defaults to None.
+            json (dict, optional): JSON data to send in the request. Defaults to None.
+
+        Returns:
+            resp: response object.
+        """
```
raphaelpclt: Same for the numpy format

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,270 @@
...
+        return resp
+
+
+    def take_measurement(self):
+        """
+        Take a measurement, save the data, and return the processed image.
+
+        Returns:
+            numpy.ndarray: Processed image data after measurement.
+        """
```
raphaelpclt: Same

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,270 @@
...
+            rotate (int): Rotation angle in degrees.
+            fliplr (bool): If True, flip the image horizontally.
+            img (numpy.ndarray): Image data to be processed.
+            mask (numpy.ndarray): Mask data to be applied.
+            wavelength (float, optional): Wavelength for scaling. Defaults to 632.8 nm.
+            create_fits (bool, optional): If True, save the processed image as a FITS file.
+
+        Returns:
+            numpy.ndarray: Processed image data.
+        """
```
raphaelpclt: Same docstring comment

catkit2/services/accufiz_interferometer_sim/accufiz_interferometer_sim.py
```diff
@@ -0,0 +1,306 @@
...
+    Generate a dummy mask as a numpy ndarray of size (h, w).
+
+    Args:
+        radius (int, optional): Proportion of the circle. Defaults to 1.
+        h (int, optional): Height of the rectangle. Defaults to 256.
+        w (int, optional): Width of the rectangle. Defaults to 256.
+
+    Returns:
+        numpy.ndarray: Dummy mask as a uint8 array.
+    """
```
raphaelpclt: Same

catkit2/services/accufiz_interferometer_sim/accufiz_interferometer_sim.py
```diff
@@ -0,0 +1,306 @@
...
+    Rotate and/or flip the image data.
+
+    Args:
+        data (numpy.ndarray): Numpy array of image data.
+        theta (int): Rotation in degrees.
+        flip (bool): If True, flip the image horizontally.
+
+    Returns:
+        numpy.ndarray: Modified image after rotation and/or flip.
+    """
```
raphaelpclt: Same numpy docstring comment

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
```diff
@@ -0,0 +1,270 @@
...
+            while self.should_be_acquiring.is_set() and not self.should_shut_down:
+                img = self.take_measurement()
+
+                has_correct_parameters = np.allclose(self.images.shape, img.shape)
+
+                if not has_correct_parameters:
+                    self.images.update_parameters('float32', img.shape, 20)
+
+                self.images.submit_data(img.astype('float32'))
+                time.sleep(0.05)
```
raphaelpclt: Just noticing that right now. Is there a specific reason you'd want a sleep here?

catkit2/services/accufiz_interferometer_sim/accufiz_interferometer_sim.py
```diff
@@ -0,0 +1,306 @@
...
+            rotate (int): Rotation angle in degrees.
+            fliplr (bool): If True, flip the image horizontally.
+            img (numpy.ndarray): Image data to be processed.
+            mask (numpy.ndarray): Mask data to be applied.
+            wavelength (float, optional): Wavelength for scaling. Defaults to 632.8 nm.
+            create_fits (bool, optional): If True, save the processed image as a FITS file.
+
+        Returns:
+            numpy.ndarray: Processed image data.
+        """
```
raphaelpclt: Docstring format 😬

catkit2/services/accufiz_interferometer_sim/accufiz_interferometer_sim.py
```diff
@@ -0,0 +1,306 @@
...
+        Simulate an HTTP POST request.
+
+        Args:
+            url (str): URL to send the POST request to.
+            data (dict, optional): Data to send in the request. Defaults to None.
+            json (dict, optional): JSON data to send in the request. Defaults to None.
+
+        Returns:
+            sim_response: Simulated response object.
+        """
```
raphaelpclt: Docstring format

catkit2/services/accufiz_interferometer_sim/accufiz_interferometer_sim.py
```diff
@@ -0,0 +1,306 @@
...
+        Simulate an HTTP GET request.
+
+        Args:
+            url (str): URL to send the GET request to.
+            params (dict, optional): Parameters for the request. Defaults to None.
+
+        Returns:
+            sim_response: Simulated response object.
+        """
```
raphaelpclt: Docstring format

raphaelpclt (changes requested):
Almost there! 
Thank you so much for all the docstring - and I feel bad requesting so many changes. But it would be nice if all the docstring would match the numpy style (it seems you have the google style). I've put one example. And you can probably set up your code editor if that can help.

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
lanemeier7: Doh! Between you and me I have an AI copilot do doc string documentation for me so no worries.

catkit2/services/accufiz_interferometer/accufiz_interferometer.py
lanemeier7: No particular reason. Especially since I've migrated to the camera interface loop approach. Good catch!

raphaelpclt (approved):
Looks good.

PR #260, opened 2024-10-25
Make DM map and command functions to outside of ServiceProxy.

ehpor:
This allows us to operate on DM commands without spinning up an entire simulator.

### DIFF
--- a/catkit2/testbed/proxies/deformable_mirror.py
+++ b/catkit2/testbed/proxies/deformable_mirror.py
@@ -91,14 +91,7 @@ class DeformableMirrorProxy(ServiceProxy):
         array
             A 1D array containing the DM command.
         """
-        try:
-            dm_map = np.broadcast_to(dm_map, self.device_actuator_mask.shape)
-        except ValueError:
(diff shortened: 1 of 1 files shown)


catkit2/testbed/proxies/deformable_mirror.py
```diff
@@ -148,20 +138,80 @@ def write_dm_shape(self, fname, dm_shape):
...
+        The path where to write the DM shape.
+    dm_shape : ndarray
+        Either a DM command or DM maps.
+    device_actuator_mask : array
+        A mask describing the physical actuators on all DMs.
+    '''
+    command_length = np.sum(device_actuator_mask)
+
+    # If we're given a DM map, convert it to a command.
+    if dm_shape.shape != (command_length,):
```
ivalaginja, line 202: Is the comma here on purpose?

ehpor: Yes, it's a comparison with a tuple. No new code, it was just moved from inside to outside the class.

ivalaginja: oh right

ivalaginja (approved):
I had hacked my way through this for THD2, thanks for actually doing this properly.

PR #263, opened 2024-11-06
Refactor synchronization structure

ehpor:
Refactor the synchronization methods inside data streams. This makes it easier to add multiple types of synchronization per operating system and to switch between them with only a few lines of code.

~~The proposed system uses the CRTP ([Curiously recurring template pattern](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern)) instead of compiler pre-directives.~~ The proposed system uses template specialization instead of compiler pre-directives. This makes it possible, at no runtime cost, to choose the synchronization method. This PR just adds the synchronization methods that were already implemented.

This PR does not intend to change any existing code, just reorder/reformat/move it to allow for a more modular structure.


raphaelpclt:
I am looking at it, but I am trying to familiarize myself with those concepts before approving

ehpor:
After looking over this PR a second time, I seem to be focussed on avoid the effects of a virtual function call (which is ~7ns). Even ignoring that this is a negligible amount of time already, this cost is paid at the entry of the function (ie. at the point in time when the function is called), not at the return of the function. As these synchronization function calls are meant to be called only when the thread is expected to wait for the frame to come in. So, _at most_ this incurs a 7ns increase in latency, in case the frame became available between the check of the data stream and the call to `Wait()`. In exchange, this PR uses CRTP which is much harder for non-experts in C++ to understand, or even read.

To conclude: this PR solves the problem with an advanced, harder to maintain language features, at the gain of 7ns of performance in the (extremely unlikely) optimum case. Therefore, I'm rewriting this PR to use normal virtual functions, which is the standard way to solve this problem. If performance is an issue (again, _highly_ unlikely), then we can revert to CRTP.

@raphaelpclt

ehpor:
I went a third way: template specialization. It's easier to reason about and it still is compile-time polymorphism. Tests pass on all OSs. Ready for test on a windows machine.

raphaelpclt:
Just compiled on HiCAT, finished with that error:
image

The server is running though

EDIT: this is also happening on develop, so not related to this PR.

raphaelpclt (dismissed):
Looks good as far as I can understand

ehpor:
@raphaelpclt That error means that, while the library was compiled correctly, it did not replace the installed version. So this was not a sufficient test. This error likely means that some catkit2 process was still running in the background. Or code completion in pycharm/vscode, sometimes.

raphaelpclt:
Finally tested, all good for me

raphaelpclt (approved):

PR #271, opened 2024-12-11
makes thorlabs_tsp01 temperature sensor heads optional per configuration

### LINKED ISSUE #270: Thorlabs tsp01 service main expects 2 external temp sensors
The Thorlabs tsp01 temperature and humidity sensor can be configured with up to 2 external temperature sensors. When the temperature sensors are not used the service logs traceback error messages. In cases where none or 1 of the external sensors are used it will log a complaint every measurement check.

https://github.com/spacetelescope/catkit2/blob/825ce48c0ac79c7c7c6faae8f5fb250187c22987/catkit2/services/thorlabs_tsp01/thorlabs_tsp01.py#L25-L44

lanemeier7:
Closes #270 by making the Thorlabs tsp01 temperature sensor probes optional via services.yml configuration.

``` yaml
tsp01_1:
  service_type: thorlabs_tsp01
  simulated_service_type: thorlabs_tsp01_sim
  requires_safety: false
  serial_number: M00797061
  num_averaging: 1
  interval: 1
  read_out_header_1: true
  read_out_header_2: false
```

### DIFF
--- a/catkit2/services/thorlabs_tsp01/thorlabs_tsp01.py
+++ b/catkit2/services/thorlabs_tsp01/thorlabs_tsp01.py
@@ -16,10 +16,12 @@ class ThorlabsTSP01(Service):
         self.serial_number = self.config['serial_number']
         self.num_averaging = self.config.get('averaging', 1)
         self.interval = self.config.get('interval', 10)
+        self.read_out_header_1 = self.config.get('read_out_header_1', True)
+        self.read_out_header_2 = self.config.get('read_out_header_2', True)
 
         self.temperature_internal = self.make_data_stream('temperature_internal', 'float64', [1], 20)
-        self.temperature_header_1 = self.make_data_stream('temperature_header_1', 'float64', [1], 20)
-        self.temperature_header_2 = self.make_data_stream('temperature_header_2', 'float64', [1], 20)
+        self.temperature_header_1 = self.make_data_stream('temperature_header_1', 'float64', [1], 20) if self.read_out_header_1 else None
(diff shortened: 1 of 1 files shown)


catkit2/services/thorlabs_tsp01/thorlabs_tsp01.py
```diff
@@ -16,10 +16,12 @@ def __init__(self):
         self.serial_number = self.config['serial_number']
         self.num_averaging = self.config.get('averaging', 1)
         self.interval = self.config.get('interval', 10)
+        self.read_out_header_1 = self.config.get('read_out_header_1', True)
+        self.read_out_header_2 = self.config.get('read_out_header_2', True)
 
         self.temperature_internal = self.make_data_stream('temperature_internal', 'float64', [1], 20)
-        self.temperature_header_1 = self.make_data_stream('temperature_header_1', 'float64', [1], 20)
-        self.temperature_header_2 = self.make_data_stream('temperature_header_2', 'float64', [1], 20)
+        self.temperature_header_1 = self.make_data_stream('temperature_header_1', 'float64', [1], 20) if self.read_out_header_1 else None
```
ehpor: Make this and the next line a normal if-statement instead of a terniary. Our variable doesn't need `None` as a default and can just not exist.

```python
if self.read_out_header_1:
    self.make_data_stream('temperature_header_1', 'float64', [1], 20)
```

ehpor (changes requested):
Small change. The ternary makes it non-trivial to read what is going on.

ehpor (approved):
I like the default of `True` too.

Issue #73, opened 2022-12-13
Clarify in installation instructions how much shared memory is needed (MacOS)


ivalaginja:
The installation instructions in the readme of catkit2 just state that the default of 4 MB of shared memory on MacOS is not enough, but not how much is recommended (see https://github.com/spacetelescope/catkit2#shared-memory-configuration).

It seems like no action is needed for MacOS Catalina. I am on MacOS Monterey however, and from what I can tell its default is still at around 4 MB. I found this set of instructions how to deal with shared memory allocation on MacOS:
http://www.spy-hill.net/help/apple/SharedMemory.html
Edit: I also found this link useful: https://www.postgresql.org/docs/8.4/kernel-resources.html

And when I run `sysctl -A | grep shm` on my MacOS 12.2 (Monterey) it returns:
```bash
kern.sysv.shmall: 1024
kern.sysv.shmmax: 4194304
kern.sysv.shmmin: 1
kern.sysv.shmmni: 32
kern.sysv.shmseg: 8
```

The current value for `kern.sysv.shmmax` suggests a max of ~4MB if I see that right, is that correct?

The link above tells me how to change my shared memory settings but I don't know at this point what the recommended settings are.

@ehpor what would you suggest I try? If whatever I try works fine, I can update the readme accordingly.

RemiSoummer:
@ivalaginja @ehpor , I'm reviving this discussion with @raphaelpclt.  
I'm running Ventura 13.6 and been working on re-installing everything from scratch (as well as working on updating install documentation).   Trying to run the GUI in simulation I have a `AttributeError: Too many open files`
which seems related to the discussion about sysctl above.  I'm investigating how to increase that, and will put that in my PR documentation when we find the solution.

ivalaginja:
@RemiSoummer you should pull @alexisyslau into the loop, she's been having exactly the same issue on her machine and has found a solution for it.

RemiSoummer:
Thanks and yes I'm interested we are going in circles for now.

RemiSoummer:
I just upgraded to Sonoma 14.4 and the problem is gone :-)  
Still interested in getting the solution documented since that might likely come back. 

For completeness this is what I have now: 

`$ sysctl -A | grep maxfiles`
```
kern.maxfiles: 245760
kern.maxfilesperproc: 122880
```
and also: 

`$ sysctl -A | grep shm`
```
kern.sysv.shmmax: 4194304
kern.sysv.shmmin: 1
[...]

PR #272, opened 2024-12-17
Add a shared-memory message broker.

ehpor:
This PR adds a `MessageBroker` class for distributing arbitrary-sized messages. Messages are put in shared memory, either CPU or GPU shared memory (the latter not implemented), and can be distributed between processes.

A `Message` contains:
1. _A topic._ We use a string, e.g. `boston_dm/correction_howfs/value` or ` boston_dm/correction_howfs/set`, which follows a hierarchical naming convention.
2. _Metadata._ We chose to use a limited number of key-value pairs. The keys are stored statically for each message, the values are stored for each message.
3. _Data._ This is an array-like structure. The shape and dtype of this array does not have to be the same from message to message.

The `MessageBroker` allows us to listen to new messages at any level in the hierarchical structure. This means that we can listen to either a single message topic or many message topics at the same time. The `MessageBroker` also allows us to send messages with any topic.

Relies on #289.


ehpor:
Performed the first benchmarks:
```
4.09438e+06 messages per second
244 ns per message
```

This is single-threaded. The biggest bottlenecks are:
* ~40%: the lookup of the topic header and event for a topic.
* ~25%: the Event structure providing notification of other processes (MacOS specific).
* ~15%: the allocators (combined freelist and pool allocator).
* ~15%: getting a timestamp.
* ~5%: misc processing.

image
(benchmark using samply)

For the topic header and event lookup for a topic: this can in principle be condensed into a single hash map/calculation. But I'll postpone this for later. This is enough for now. All low-hanging fruit has been taken care of.

ehpor:
When walking home, I realized that the subtopics are very predictable. Doing a full lookup every subtopic might be avoidable with clever ordering. This is also work for the future, if required.

ehpor:
I'm having some problems with the Python bindings. The polymorphism for Memory, SharedMemory and LocalMemory is not working as expected. I don't really understand why not. Pybind11 knows about the inheritance structure and should upcast the shared_ptr<LocalMemory> to a shared_ptr<Memory> just fine.

ehpor:
I've moved to a cheaper memory allocator, based on a thread-local pool fed by a buddy allocator. This is the final iteration of the allocator. I've also moved to a single Event structure for signaling.

New benchmarks:
```
5.54217e+06 messages per second
180 ns per message
```
Distributed as follows:
* 10% generating UUIDs
* 23% allocating and deallocating
* 21% getting topic header
* 7% getting a timestamp
* 4% looping through subtopics
* 4% signaling the event
* 3% updating the frame counter per topic
* 9% updating the framerates
* 19% misc other stuff

ehpor:
This is a monster PR, and I'm sorry for that. The full functionality is not complete. No more functional changes. I'll only be making docstring additions and code style clarifications / minor bug fixes if I see them.

## For reviewers

Due to the size of this PR, and it being almost completely written in C++, I'm not expecting a deep dive into the C++ code. Not even a cursory glance. I am however expecting a _full and thorough review_ on the Python test code that tests the C++ code. I want to make sure the interface there is correct, as expected and sane. This includes:
* [ ] `tests/test_event.py`
* [ ] `tests/test_message_broker.py`
* [ ] `tests/test_uuid.py`

## Diagrams

Some high-level diagrams of the interface / high-level sketches of the inner workings.

The final intended use-case for the message broker:
image

Message contents:
image

Temporal flow of messages in an AO system:
image

Internals of the message broker:
image

(available here: https://excalidraw.com/#room=0400418253d5feb40532,MArWGWNpNYY_RhMSaQTXjA )

ivalaginja:
@ehpor thank you for the extensive visuals, they help a lot in understanding what's going on!

Could you maybe give us a very top-level application case for the message broker? I see that this facilitates messaging between separate machines - how us the user intended to interact with this new functionality, is it optional to tap into that or is it something we actively engage with every time when writing/using code?

ehpor:
@ivalaginja A short summary first, then detailed descriptions of each of the points.

# Ultimate goal

The ultimate goal is to have the message broker replace the current split between data streams (= via shared memory) and commands/properties (= via TCP). Currently, users are unable to introspect the communication that is happening via TCP. This PR is intending to provide a unified way to communicate with/between services, while still using the current API of properties, commands and streams. (Perhaps removing datastreams at some later point, since properties will be just as fast.) While the user could tap into this communication stream directly, it is not required unless you want to do special things. Additionally, it would Furthermore, it would make spreading catkit2 over multiple computers seamless.

This being said however, if you are interested in catkit2 internals, understanding the message broker interface (i.e. what is tested in the `test_*.py` files) would be important to understand its strengths / limitations.

The transition to message broker communication would be gradual, with a service property/command/datastream wrapping the message stream interaction, and optional features usable after that transition. I'm not yet sure if the full transition can be done backwards-compatible, but that's what I'm aiming for.

# Getting and setting styles and introspection

There are currently four different styles of setting/getting "things" on a service in catkit2. I'll give examples for each one.

1. A property. This communication happens fully via TCP. For an external process, other than polling, there is no way for it to "see" if/when a property changes value. We tend to use this for properties which are rarely gotten/set, like the channel names on a DM or the names of the checked safeties on the safety monitor.
2. A datastream-backed property. Here, the set happens via TCP, but the get happens via a datastream. This has the advantage of having datastreams, without the complication of writing one explicitly. However, setting the property is still invisible to external processes. Furthermore, it is still impossible to be notified / wait on the stream for changes in the underlying service value. This means that to check for changes, we're still polling, even though the polling itself doesn't disturb the service itself anymore.
3. A datastream. Both setting and getting happens via shared memory. This seems like a great solution, but there is a difference between _requesting_ the underlying service value to change vs. _distributing_ the current underlying value of the service. An example here: for DMs, the `correction_howfs` stream is to request changes to the DM surface. There is no way for us to figure out when the surface is actually applied to the DM. On the other hand, the `total_surface` stream is there to tell you what the current DM command is. While setting something on the `total_surface` stream is not prohibited, it will have no effect on the actual command on the DM and you'd be lying to external processes. To summarize: it seems like there are two "kinds" of streams: a get-only and set-only. (Which is strange by itself, since usually, programmers expect either read-only or read-write properties, not read-only or write-only properties.) The fourth style fixes this.
4. Two datastreams. Here we separate out two datastreams explicitly. We usually use this style when operations take a long time and we _need_ to know when a change finished. Because this is not enforced in catkit2, we confusingly use different nomenclature for the two streams in different services. For example, the Newport XPS Q8 uses `{axis_name}_command` for requesting changes and `{axis_name}_current` for the actual location of the motor. The Thorlabs MF101 uses `commanded_position` and `current_position`. And the Newport picomotor uses `{axis_name}_command` and `{axis_name}_current_position`. While this style is obviously the most comprehensive, it is also the most expensive, see the next section.

A message-broker-based solution however can use two message topics per property, formalized by the implementation of a service property. The implementation in this PR does not specify the convention for this. But generally, it would be something like the following. A flip mount service would listen for `seal_windows/source_flip_mount/position/set` messages. When it receives one, it moves the flip mount to the requested position, then publish a `seal_windows/source_flip_mount/position/get` message to indicate that the state of the flip mount has changed. A similar procedure would be the case for `seal_linux/boston_dm/correction_howfs/set` and the corresponding `*/get` message. All of this is abstracted away behind a `testbed.source_flip_mount.position = 'in_beam'` accessor on the service. No need to interact with a message broker yourself. Note, I'm not sure on the syntax for that yet, since I want to support both blocking and non-blocking setters. The same for `testbed.boston_dm.correction_howfs = new_command`. Note that any external process can observe what is happening by subscribing to messages: `subscription = testbed.boston_dm.correction_howfs.subscribe()` and then `for cmd in subscription:` or something. Again, not sure about the syntax yet.

# Resource consumption

We could in principle rewrite each property to use style 4, having two datastreams each. The issue that properties allow arbitrarily-sized data, while datastreams require fixed sized data and expensive changes to data size, this could be worked around by optimizing datastreams themselves.

However, this is ignoring one important property: each style differs in the number of used shared memory segments. The number of open memory segments (= file descriptors) is limited, so you can only access a limited number of properties on a particular process. We have experienced issues with this in the past (e.g. https://github.com/spacetelescope/catkit2/issues/73#issuecomment-2062101479 and others). Respectively, style 1, 2, 3, 4 requires 0, 3, 3, 6 open files per property (1 header shared memory, 1 data shared memory, 1 semaphore, all per stream). For the bigger services, that is limiting. While the limit on the number of open files can be increased, as has been our fix so far, this indicates that we're doing something wrong here.

On the other hand, a message-broker-based solution would only require access to two shared memory segments and accompanying synchronization structures, independent of the number of properties / streams / services your process interacts with. 

# Multi-machine testbeds

While the message broker itself uses shared memory, and is therefore not capable of communicating between machines, it provides a clean interface to do so, a single end-point to listen to for anything that happens on that machine. Therefore, a _future_ PR will write a communication using ZMQ sockets to communicate messages from the message broker across machines and inject those messages into the local message broker. In that way, you, as a user, won't even notice that a service is located on a different computer. All communication is abstracted away into messages and you don't care how those messages get to their destination. (At a certain point, you would care of course. Being communication between machines increases latency. So I would not recommend having a camera, DM and wavefront controller on different machines, if you're building a real AO system. But for initial tests / integration / testbed usage, it would be sufficient.)

Note that this _future_ communication code is only written once. Currently, if we have a service operating on another machine, we have custom ZMQ communication and a custom protocol between those computers for that specific service. For example, the flip mounts on SEAL are on `seal_windows`, separate from the main control computer `seal_linux`. On HiCAT however, the flip mounts run on the same computer. This means that, for SEAL, we need separate service code to operate the flip mounts compared to HiCAT. The same for our IrisAO, our K-cube motor controller and our light source. Furthermore, while HiCAT has had lots of problems with the flip mounts and other minor USB devices, moving those to a separate PC would incur a significant development effort, meaning that we deal with the inconveniences and never moved them to a separate machine. Having the communication code abstracted away behind a message broker means that we can move devices between computers with just config file changes, perhaps just even one line, and only have to deal with driver installation issues on the other machine.

tests/test_message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+    broker = MessageBroker.create(header, [block])
+    yield broker
+
+def test_message_subscription(broker):
+    topic = "test_message_subscription"
+
+    subscription_newest = broker.subscribe(topic, mode=MessageSubscriptionMode.NewestOnly)
+    subscription_sequential = broker.subscribe(topic, mode=MessageSubscriptionMode.Sequential)
+
+    for i in range(3):
```
ivalaginja, line 19: Minor: Would it be possible to use another array content than 0,1,2 ? These are currently identical to the message IDs, so it took me two looks to identify in which case I am expecting what.

tests/test_message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+    assert retrieved_message.payload.itemsize == arr.itemsize
+    assert retrieved_message.payload.ndim == len(shape)
+    assert retrieved_message.topic == topic
+
+def test_message_broker_publish(broker):
+    topic = "test_message_broker_publish"
+    data = b'hello world'
+    arr = np.frombuffer(data, dtype='uint8')
+
+    broker.publish_data(topic, data)
```
ivalaginja, line 84: How is this different from `publish_message()`? Also, what is data? Is it a synonym for payload? Trying to relate things to the figures in https://github.com/spacetelescope/catkit2/pull/272#issuecomment-2824721104.

tests/test_message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+    topic = "test_message_broker_trace_id"
+    data = b'hello world'
+
+    broker.publish_data(topic, data)
+    message_1 = broker.get_newest_message(topic)
+
+    broker.publish_data(topic, data * 2)
+    message_2 = broker.get_newest_message(topic)
+
+    assert str(message_2.trace_id) != str(message_1.trace_id)
```
ivalaginja, line 100: What are the differences between trace_id, message_id and Partial frame ID?

tests/test_event.py
```diff
@@ -0,0 +1,46 @@
...
+    signaled = threading.Event()
+    waiting = threading.Event()
+
+    condition = [0]
+
+    thread = threading.Thread(target=event_wait, args=(event, signaled, waiting, condition))
+    thread.start()
+
+    waiting.wait(1)
+    assert waiting.is_set()
```
ivalaginja: If this times out without a signal, shouldn't `waiting.is_set()` return `False`?

ivalaginja (commented):
I am slowly getting accustomed to this new concept but it's taking some time. I have a first round of questions that I was hoping to get answers to before being able to wrap this up on my side.

catkit_core/MessageBroker.cpp
```diff
@@ -0,0 +1,919 @@
...
+	auto message = PrepareMessage(topic, data_size, trace_id, memory_block_id);
+
+	// Copy the data to the payload.
+	std::memcpy(message.m_Payload, data, data_size);
+
+	// Set the array information.
+	ArrayInfo &array_info = message.m_Header->payload_info.array_info;
+
+	array_info.data_type = 'u';
+	array_info.byte_order = '=';
```
alexisyslau, line 534: Most of the CPUs and GPUs should be little-endian so using the native order should be ok normally. But is it better to force it to be little-endian `<`? This will make sure we always interpret the array in the same way.

tests/test_message_broker.py
alexisyslau: @ehpor Please correct me if I am wrong. (and yes, I did dive a bit into the C++ code to see what's going on) 

The difference is that `publish_message` already has a message object ready to be published to the broker, but that is not the case here. In this example, we only have `topic` and `data`, without any message object being created yet.

Without publishing a message, the broker has no access to any information.
So `publish_data()` needs to do several things:
- Assign a trace_id, which is needed to trace our message.
- Prepare a message, meaning it creates a message object (Header and Payload, see the message content fig. ) ready for publication.
- Set the data information for the message, so it is complete and ready to be published.
-  Publish the message, making it available to the broker.

Hence, `assert (retrieved_message.payload == arr).all()` should pass. `payload` stores the data of the message.

tests/test_event.py
ehpor: Yes, this is checking that the wait didn't time out. I could also do: 
```python
assert waiting.wait(1)
```

I'll add comments.

tests/test_message_broker.py
ehpor: Maybe make the data `2 * i` to minimize the confusion?

tests/test_message_broker.py
ehpor: @ivalaginja Data is not a synonym for payload. A payload is part of a message. Data is not part of a message. But a payload is data.

The reason for this function is similar to the `Datastream.submit_data()` vs. `Datastream.request_new_frame()` & `Datastream.submit_frame()`. The first is a convenience function for the latter two. Maybe my naming can be improved?

@alexisyslau Exactly. Unless you're hyperoptimizing, most people are not using the `Datastream.request_new_frame()` and `Datastream.submit_frame()` separately. Using those allows, in some cases, to remove a copy of the image. But most people will be using the `Datastream.submit_data()`, which combines the two, and the memory copy in between.

The same thing is true here. I'm expecting very little use of `MessageBroker.prepare_message()` & `MessageBroker.publish_message()` in favor of `MessageBroker.publish_data()` which combines the two and the memory copy in between. However, similar to the case of Datastreams, I want it to be available when people need it.

tests/test_message_broker.py
ehpor: A trace ID is set for each new information. It is typically assigned to the original data that came in, ie. the source of the "information". Eg. a camera service would create new trace IDs for each camera image. But a WFS reconstructor service would retain the same trace ID of the image that it computed the reconstruction for. Therefore, by using the trace ID, we can trace the flow of that camera information through the entire system. The technical term for this is "distributed tracing".

A payload ID is unique per data. In case partial messages are submitted (ie. pixel streaming), the payload for these messages will be the same, so they retain the same payload ID. A payload ID identifies a unique array.

A message/frame ID is sequential numbering per topic. The trace ID and payload ID are not. It allows us to figure out whether we missed frames, and to request the previous/next message for a particular message.

A partial frame ID is a numbering within a message. It increments sequentially, starting from zero. It gets incremented each time a partial message is submitted. I didn't test the partial message generation in this PR. It probably needs more work.

catkit_core/MessageBroker.cpp
ehpor: Since the data type here is a byte (`u`), which is one byte long, the byte order doesn't matter. But, I'm not sure why I chose `=` compared to `@`, which are native byte order and native alignment, and native byte order and no alignment respectively. Again, no difference in this case, but native everything should be the default in other parts of the code, e.g. when submitting an array.

tests/test_message_broker.py
ehpor: Changed the data to `i + 10`, so it's now different than the message IDs.

tests/test_event.py
ehpor: Added more comments to the test.

tests/test_message_broker.py
ivalaginja: I see! Thanks to both of you! I think the naming is fine, especially when you trace it back to the data streams. Makes sense!

tests/test_event.py
ivalaginja: Oh I had inverted the purpose of the test in my head. Makes sense! Thanks!

tests/test_message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+    header = LocalMemory.create(1024 * 1024 * 512)
+    block = LocalMemory.create(1024 * 1024 * 1024)
+
+    broker = MessageBroker.create(header, [block])
+    yield broker
+
+def test_message_subscription(broker):
+    topic = "test_message_subscription"
+
+    subscription_newest = broker.subscribe(topic, mode=MessageSubscriptionMode.NewestOnly)
```
ivalaginja, line 16: New question: why do we need to subscribe to the broker, and then work through the subscription? Intuitively, I would have asked the broker directly for a next message, next message ID etc. Is it only to be able to define the subscription mode or is there more behind it?

tests/test_message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+
+    assert message_1 is not None
+    assert message_2 is not None
+    assert message_3 is not None
+
+    assert message_1.payload[0] == 12
+    assert message_2.payload[0] == 10
+    assert message_3.payload[0] == 11
+
+    assert subscription_newest.try_get_next_message() is None
```
ivalaginja, line 40: So this means once I get a message through a subscription, that message does not exist anymore? Would make sense, just double-checking - and trying to find the correct vocabulary for this ^^

tests/test_message_broker.py
```diff
@@ -0,0 +1,105 @@
...
+
+    m = subscription_sequential.try_get_next_message()
+    assert m is not None
+    assert m.payload[0] == 12
+
+    assert subscription_sequential.try_get_next_message() is None
+
+    subscription_newest2 = broker.subscribe(topic, starting_frame_id=1, mode=MessageSubscriptionMode.NewestOnly)
+    assert subscription_newest2.next_message_id == 2
```
ivalaginja, line 49: Why is this 2 if you give it a starting ID of 2? Does it not give you the same one? The sequential subscription below shows the behavior I expect, so why does the "newest" subscription mode automatically up the message id?

ivalaginja (commented):
I left a couple more questions about some details, but over all things make a lot more sense now. I also still need to run these tests.

ivalaginja:
@ehpor I am unsure what I am doing wrong. I am trying to reinstall catkit2 so that I can test the tests, but the installation fails with the below error message. I am slowly acquiring PTSD from Cmake errors 😬 

```bash
(catkit2) maintenances-MacBook-Pro-6:catkit2 ilaginja$ pip install -e .
Obtaining file:///Users/ilaginja/repos/catkit2
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Installing backend dependencies ... done
  Preparing editable metadata (pyproject.toml) ... done
Building wheels for collected packages: catkit2
  Building editable for catkit2 (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building editable for catkit2 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [41 lines of output]
      2025-05-07 09:28:20,644 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake -E capabilities
      2025-05-07 09:28:20,674 - scikit_build_core - INFO - CMake version: 3.29.3
      *** scikit-build-core 0.11.2 using CMake 3.29.3 (editable)
      2025-05-07 09:28:20,724 - scikit_build_core - INFO - Build directory: /private/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmptmubhn3l/build
      *** Configuring CMake...
      2025-05-07 09:28:20,765 - scikit_build_core - INFO - RUN: ninja --version
      2025-05-07 09:28:21,413 - scikit_build_core - INFO - Ninja version: 1.11.1
      2025-05-07 09:28:21,415 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake -S. -B/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmptmubhn3l/build -DCMAKE_BUILD_TYPE:STRING=Release -C/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmptmubhn3l/build/CMakeInit.txt -DCMAKE_INSTALL_PREFIX=/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmptmubhn3l/wheel/platlib -DCMAKE_MAKE_PROGRAM=ninja
      loading initial cache file /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmptmubhn3l/build/CMakeInit.txt
      -- The C compiler identification is AppleClang 13.0.0.13000027
      -- The CXX compiler identification is AppleClang 13.0.0.13000027
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/clang - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/clang++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      CMake Error at catkit_core/CMakeLists.txt:41 (find_package):
        By not providing "Findcppzmq.cmake" in CMAKE_MODULE_PATH this project has
        asked CMake to find a package configuration file provided by "cppzmq", but
        CMake did not find one.
      
        Could not find a package configuration file provided by "cppzmq" with any
        of the following names:
      
          cppzmqConfig.cmake
          cppzmq-config.cmake
      
        Add the installation prefix of "cppzmq" to CMAKE_PREFIX_PATH or set
        "cppzmq_DIR" to a directory containing one of the above files.  If "cppzmq"
        provides a separate development package or SDK, be sure it has been
        installed.
      
      
      -- Configuring incomplete, errors occurred!
      
      *** CMake configuration failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building editable for catkit2
Failed to build catkit2
ERROR: Could not build wheels for catkit2, which is required to install pyproject.toml-based projects
(catkit2) maintenances-MacBook-Pro-6:catkit2 ilaginja$ 

```

ehpor:
For me, this has always been a problem of not activating the environment, or, worse, conda telling me the environment is activated while it is not. After deactivating the environment, then reactivating it again, it works for me. Another check is to see if cppzmq is installed `conda list cppzmq`. It should be.

FYI: I've been becoming so fed up with this, that there is #312 which removes cppzmq as dependency (not the root cause btw), and #276 (for doing a correct build of the library).

ivalaginja:
... it actually seems like I don't have cppmq installed? No output when I run `conda list cppzmq`, or `conda list |grep cpp` (tried the latter just in case I mistyped).

tests/test_message_broker.py
ehpor: One big problem currently with datastreams is that the next frame ID is linked to your opened datastream. Furthermore, service proxies cache the datastream object, since closing and reopening is an expensive operation and so we are minimizing that. This means that if you do `stream.get_next_frame()` on the same stream, given to you on the same service, the threads will compete for the next frame. Ie. the threads are bound to the same subscription.

By splitting the subscription up from the message broker / datastream, you avoid this problem. The next frame id to read is stored in the subscription, where it's supposed to be, rather than in the message broker / datastream. This was an oversight when developing catkit2 datastreams, and I'm kinda surprised that we haven't run into bugs of this kind so far.

ivalaginja:
I'll just install it manually. God knows on what outdated env file I created this environment.

ivalaginja:
Uuughh. Time for a completely new environment from scratch?

```bash
(catkit2) maintenances-MacBook-Pro-6:catkit2 ilaginja$ pip install -e .                                        
Obtaining file:///Users/ilaginja/repos/catkit2                                                                 
  Installing build dependencies ... done                                                                       
  Checking if build backend supports build_editable ... done                                                   
  Getting requirements to build editable ... done                                                              
  Installing backend dependencies ... done                                                                     
  Preparing editable metadata (pyproject.toml) ... done                                                        
Building wheels for collected packages: catkit2
  Building editable for catkit2 (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building editable for catkit2 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [37 lines of output]
      2025-05-07 09:42:40,744 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake -E capabilities
      2025-05-07 09:42:40,772 - scikit_build_core - INFO - CMake version: 3.29.4
      *** scikit-build-core 0.11.2 using CMake 3.29.4 (editable)
      2025-05-07 09:42:40,820 - scikit_build_core - INFO - Build directory: /private/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmpbgzqkwlc/build
      *** Configuring CMake...
      2025-05-07 09:42:40,855 - scikit_build_core - INFO - RUN: ninja --version
      2025-05-07 09:42:41,118 - scikit_build_core - INFO - Ninja version: 1.11.1
      2025-05-07 09:42:41,120 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake -S. -B/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmpbgzqkwlc/build -DCMAKE_BUILD_TYPE:STRING=Release -C/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmpbgzqkwlc/build/CMakeInit.txt -DCMAKE_INSTALL_PREFIX=/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmpbgzqkwlc/wheel/platlib -DCMAKE_MAKE_PROGRAM=ninja
      loading initial cache file /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmpbgzqkwlc/build/CMakeInit.txt
      -- The C compiler identification is AppleClang 13.0.0.13000027
      -- The CXX compiler identification is AppleClang 13.0.0.13000027
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/clang - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/clang++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      CMake Error at catkit_core/CMakeLists.txt:45 (find_package):
        Could not find a package configuration file provided by "Eigen3" with any
        of the following names:
      
          Eigen3Config.cmake
          eigen3-config.cmake
      
        Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
        "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
        provides a separate development package or SDK, be sure it has been
        installed.
      
      
      -- Configuring incomplete, errors occurred!
      
      *** CMake configuration failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building editable for catkit2
Failed to build catkit2
ERROR: Could not build wheels for catkit2, which is required to install pyproject.toml-based projects
(catkit2) maintenances-MacBook-Pro-6:catkit2 ilaginja$ 
```

tests/test_message_broker.py
ehpor: Yep, a single subscription can only access a certain message once. The `try_get_next_frame()` function is there for datastreams as well, but hidden behind a `timeout=0`, which reads badly.

So, there is `get_next_message()` which gets the next message in your subscription, according to your `MessageSubscriptionMode`, which will wait for that message if it's not there yet or return immediately if it is. And we have `try_get_next_message()` which doesn't wait and will return None if it's not available yet.

ivalaginja:
@ehpor I really don't know what I am doing wrong. I ran `conda env update --file environment.yml` on this branch, then `pip install -e .` and I still can't install things.

```bash
(catkit2) maintenances-MacBook-Pro-6:catkit2 ilaginja$ pip install -e .
Obtaining file:///Users/ilaginja/repos/catkit2
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Installing backend dependencies ... done
  Preparing editable metadata (pyproject.toml) ... done
Building wheels for collected packages: catkit2
  Building editable for catkit2 (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building editable for catkit2 (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [94 lines of output]
      2025-05-07 09:49:08,924 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake -E capabilities
      2025-05-07 09:49:08,975 - scikit_build_core - INFO - CMake version: 3.27.4
      *** scikit-build-core 0.10.7 using CMake 3.27.4 (editable)
      2025-05-07 09:49:09,068 - scikit_build_core - INFO - Build directory: /private/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build
      *** Configuring CMake...
      2025-05-07 09:49:09,130 - scikit_build_core - INFO - RUN: ninja --version
      2025-05-07 09:49:09,380 - scikit_build_core - INFO - Ninja version: 1.11.1
      2025-05-07 09:49:09,384 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake -S. -B/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build -DCMAKE_BUILD_TYPE:STRING=Release -C/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/CMakeInit.txt -DCMAKE_INSTALL_PREFIX=/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/wheel/platlib -DCMAKE_MAKE_PROGRAM=ninja
      loading initial cache file /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/CMakeInit.txt
      -- The C compiler identification is AppleClang 13.0.0.13000027
      -- The CXX compiler identification is AppleClang 13.0.0.13000027
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/gcc - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/g++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      -- Found nlohmann_json: /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/cmake/nlohmann_json/nlohmann_jsonConfig.cmake (found version "3.9.1")
      -- Found Protobuf: /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/libprotobuf.dylib (found version "3.20.3")
      CMake Warning (dev) at /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/python3.7/site-packages/pybind11/share/cmake/pybind11/FindPythonLibsNew.cmake:101 (message):
        Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
        are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
        the cmake_policy command to set the policy and suppress this warning, or
        preferably upgrade to using FindPython, either by calling it explicitly
        before pybind11, or by setting PYBIND11_FINDPYTHON ON before pybind11.
      Call Stack (most recent call first):
        /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/python3.7/site-packages/pybind11/share/cmake/pybind11/pybind11Tools.cmake:50 (find_package)
        /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/python3.7/site-packages/pybind11/share/cmake/pybind11/pybind11Common.cmake:228 (include)
        /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/python3.7/site-packages/pybind11/share/cmake/pybind11/pybind11Config.cmake:250 (include)
        catkit2/CMakeLists.txt:10 (find_package)
      This warning is for project developers.  Use -Wno-dev to suppress it.
      
      -- Found PythonInterp: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/python3.7 (found suitable version "3.7.16", minimum required is "3.7")
      -- Found PythonLibs: /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/libpython3.7m.dylib
      -- Performing Test HAS_FLTO_THIN
      -- Performing Test HAS_FLTO_THIN - Success
      -- Found pybind11: /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/python3.7/site-packages/pybind11/include (found version "2.13.6")
      -- Found nlohmann_json: /Users/ilaginja/opt/miniconda3/envs/catkit2/lib/cmake/nlohmann_json/nlohmann_jsonConfig.cmake (found suitable version "3.9.1", minimum required is "3.2.0")
      -- Configuring done (4.4s)
      -- Generating done (0.1s)
      -- Build files have been written to: /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build
      *** Building project with Ninja...
      2025-05-07 09:49:13,954 - scikit_build_core - INFO - RUN: /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/cmake --build /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build -v
      Change Dir: '/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build'
      
      Run Build Command(s): ninja -v
      [1/61] cd /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit2 && /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/protoc --proto_path=/Users/ilaginja/repos/catkit2/catkit2/../proto/ --python_out=/Users/ilaginja/repos/catkit2/catkit2/./proto/ /Users/ilaginja/repos/catkit2/catkit2/../proto/core.proto /Users/ilaginja/repos/catkit2/catkit2/../proto/logging.proto /Users/ilaginja/repos/catkit2/catkit2/../proto/service.proto /Users/ilaginja/repos/catkit2/catkit2/../proto/testbed.proto /Users/ilaginja/repos/catkit2/catkit2/../proto/tracing.proto
      [2/61] cd /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core && /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/protoc --cpp_out /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen/ -I /Users/ilaginja/repos/catkit2/catkit_core -I /Users/ilaginja/repos/catkit2/proto /Users/ilaginja/repos/catkit2/proto/core.proto
      [3/61] cd /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core && /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/protoc --cpp_out /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen/ -I /Users/ilaginja/repos/catkit2/catkit_core -I /Users/ilaginja/repos/catkit2/proto /Users/ilaginja/repos/catkit2/proto/service.proto
      [4/61] cd /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core && /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/protoc --cpp_out /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen/ -I /Users/ilaginja/repos/catkit2/catkit_core -I /Users/ilaginja/repos/catkit2/proto /Users/ilaginja/repos/catkit2/proto/tracing.proto
      [5/61] cd /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core && /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/protoc --cpp_out /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen/ -I /Users/ilaginja/repos/catkit2/catkit_core -I /Users/ilaginja/repos/catkit2/proto /Users/ilaginja/repos/catkit2/proto/logging.proto
      [6/61] cd /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core && /Users/ilaginja/opt/miniconda3/envs/catkit2/bin/protoc --cpp_out /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen/ -I /Users/ilaginja/repos/catkit2/catkit_core -I /Users/ilaginja/repos/catkit2/proto /Users/ilaginja/repos/catkit2/proto/testbed.proto
      [7/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/LocalMemory.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/LocalMemory.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/LocalMemory.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/LocalMemory.cpp
      [8/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/HostName.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/HostName.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/HostName.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/HostName.cpp
      [9/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/BuddyAllocator.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/BuddyAllocator.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/BuddyAllocator.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/BuddyAllocator.cpp
      [10/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Log.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Log.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Log.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Log.cpp
      [11/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Event.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Event.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Event.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Event.cpp
      [12/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/HybridPoolAllocator.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/HybridPoolAllocator.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/HybridPoolAllocator.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/HybridPoolAllocator.cpp
      [13/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/HashMap.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/HashMap.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/HashMap.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/HashMap.cpp
      [14/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/LogConsole.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/LogConsole.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/LogConsole.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/LogConsole.cpp
      [15/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/PoolAllocator.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/PoolAllocator.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/PoolAllocator.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/PoolAllocator.cpp
      [16/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/LogFile.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/LogFile.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/LogFile.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/LogFile.cpp
      [17/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Client.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Client.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Client.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Client.cpp
      [18/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Command.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Command.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Command.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Command.cpp
      [19/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/ServiceState.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/ServiceState.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/ServiceState.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/ServiceState.cpp
      [20/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/SharedMemory.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/SharedMemory.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/SharedMemory.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/SharedMemory.cpp
      [21/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/DataStream.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/DataStream.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/DataStream.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/DataStream.cpp
      [22/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Server.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Server.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Server.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Server.cpp
      [23/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Property.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Property.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Property.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Property.cpp
      [24/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Timing.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Timing.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Timing.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Timing.cpp
      [25/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Shareable.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Shareable.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Shareable.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Shareable.cpp
      [26/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/MessageBroker.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/MessageBroker.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/MessageBroker.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/MessageBroker.cpp
      [27/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o -MF catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o.d -o catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o -c /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/proto/core.pb.cc
      FAILED: catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o
      /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o -MF catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o.d -o catkit_core/CMakeFiles/catkit_core.dir/proto/core.pb.cc.o -c /var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/proto/core.pb.cc
      clang: error: no such file or directory: '/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/proto/core.pb.cc'
      clang: error: no input files
      [28/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Util.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Util.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Util.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Util.cpp
      [29/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/LogForwarder.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/LogForwarder.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/LogForwarder.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/LogForwarder.cpp
      [30/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Tensor.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Tensor.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Tensor.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Tensor.cpp
      [31/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Uuid.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Uuid.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Uuid.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Uuid.cpp
      [32/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Tracing.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Tracing.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Tracing.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Tracing.cpp
      [33/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/ServiceProxy.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/ServiceProxy.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/ServiceProxy.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/ServiceProxy.cpp
      [34/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Types.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Types.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Types.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Types.cpp
      [35/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/Service.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/Service.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/Service.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/Service.cpp
      [36/61] /Library/Developer/CommandLineTools/usr/bin/g++ -DPROTOBUF_USE_DLLS -I/var/folders/28/w2xb9gbd7_g3wxh48m73qv840000gp/T/tmp0xk3ekoc/build/catkit_core/gen -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include -isystem /Users/ilaginja/opt/miniconda3/envs/catkit2/include/eigen3 -O3 -DNDEBUG -std=gnu++17 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk -fPIC -g -MD -MT catkit_core/CMakeFiles/catkit_core.dir/TestbedProxy.cpp.o -MF catkit_core/CMakeFiles/catkit_core.dir/TestbedProxy.cpp.o.d -o catkit_core/CMakeFiles/catkit_core.dir/TestbedProxy.cpp.o -c /Users/ilaginja/repos/catkit2/catkit_core/TestbedProxy.cpp
      ninja: build stopped: subcommand failed.
      
      
      *** CMake build failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building editable for catkit2
Failed to build catkit2
ERROR: Could not build wheels for catkit2, which is required to install pyproject.toml-based projects
```

tests/test_message_broker.py
ehpor: `MessageSubscriptionMode.Sequential` will try to give you messages without dropped frames, even if there are newer messages available. `MessageSubscriptionMode.NewestOnly` will instead drop frames to give you the most recent update. In this case, `subscription_newest2` is trying to read frame 1, but sees that frame 2 is already there, so it will skip frame 1 and give you frame 2. `subscription_sequential2` will _not_ skip frames, and so will give you frame 1 first, then frame 2 afterwards (untested here, because that was already tested a few asserts above).

ivalaginja: (Potentially stupid) follow-up question: How come there is frame 2 already there?

ehpor:
@ivalaginja That looks like the same error that @brynickson had during the hackathon. I don't remember how she fixed that or what the problem was. I do remember if felt like magic.

ehpor:
And yes, protobuf is also something that I intend to rip out after the message broker transition. We will not need it anymore after that.

ivalaginja:
> @ivalaginja That looks like the same error that @brynickson had during the hackathon. I don't remember how she fixed that or what the problem was. I do remember if felt like magic.

@raphaelpclt @brynickson any pointers on how this was fixed?

ivalaginja:
> > @ivalaginja That looks like the same error that @brynickson had during the hackathon. I don't remember how she fixed that or what the problem was. I do remember if felt like magic.
> 
> @raphaelpclt @brynickson any pointers on how this was fixed?

@ehpor I am now told by @raphaelpclt that you were actually working through this with @brynickson, who is currently out of office ^^

raphaelpclt:
> > > @ivalaginja That looks like the same error that @brynickson had during the hackathon. I don't remember how she fixed that or what the problem was. I do remember if felt like magic.
> > 
> > 
> > @raphaelpclt @brynickson any pointers on how this was fixed?
> 
> @ehpor I am now told by @raphaelpclt that you were actually working through this with @brynickson, who is currently out of office ^^

From what I remember it was a compiling tools issue - a package needing to be updated. Maybe cmake?

ivalaginja:
Ha. @raphaelpclt and I found a solution: upgrade cmake manually to version 3.31.2

Installation worked now. Tests tbd.

ivalaginja (approved):
The python test code makes sense to me, and the tests pass on my MacOs 12.2. Note that I run a conda environment with Python 3.9.12 (THD2 default) and I had to upgrade cmake to 3.31.2.

If you could still answer my [last question](https://github.com/spacetelescope/catkit2/pull/272#discussion_r2077024399) that'd be great, but that's just for my own understanding anyway.

I guess we better hold off of merging until we get more reviews in?

tests/test_message_broker.py
ehpor: We submitted three frames to the message broker: frame 0, 1 and 2.

ehpor: Or I'm misunderstanding the question.

ehpor:
Yep, forgot to show my testing:

* Tested on Windows 11, AMD 5900X + benchmarks, Python 3.7
* Tested on Ubuntu 20.04, AMD Threadripper 5995WX + benchmarks, Python 3.8 & 3.11
* Tested on MacOS Sonoma 14.7, Macbook M2 Pro + benchmarks, Python 3.8, 3.11

There are two PRs open (#310 & #311), which reference this one, that make improvements / bugfixes. And a third one on the way working on network communications between message brokers (introducing a ShardedMessageBroker class).

tests/test_message_broker.py
ivalaginja: Submitted where? I only see them submitted at the very top, and we access them all later on, so they shouldn't be there anymore, no? Or does a new subscription have access to all the messages?

ivalaginja:
@ehpor btw, does this PR require rewriting some of the catkit2 docs? Or does this touch all our unwritten docs only? ^^

ehpor:
Only touches the unwritten part of the docs 😭

tests/test_message_broker.py
ehpor: Messages are not consumed. You can access historical data as well. The last N messages are kept for each topic (currently N = 32), similar to how the last N frames are kept for datastreams.

Sorry, I should have said that earlier. This is different than Apacha Kafka or Aeron, if you're familiar with those. In those frameworks, messages are only delivered once, and only accessible if published after you created your subscription. In our message broker, messages are delivered once and then can be accessed later, kinda like a time series database. I'm not fully convinced about the latter capability, since I don't think we _ever_ used it on HiCAT. And it complicates the networking interface that I'm building now.

ehpor:
Merging at the end of today @RemiSoummer, unless you have objections.

tests/test_message_broker.py
ivalaginja: Ah gotcha!! I don't know those in particular, but that's the type of systems I was in touch with previously. Makes sense, thanks a lot for the effort you put into all these explanations!

RemiSoummer:
@ehpor I haven't had a chance to even start looking at it, so go ahead if you need to merge it.

PR #274, opened 2024-12-20
Adding instructions for Apple Silicon

RemiSoummer:
I recently re-installed from scratch on new Mac et realized instructions need to be updated. 
This will go away once we can upgrade python version on hicat project side, we are limited by vendors SDK and stuck at 3.7 at the moment.


raphaelpclt (approved):
Ok

PR #280, opened 2025-02-02
Add trace intervals to new DM services from #156

### LINKED ISSUE #262: Add trace intervals to new DM classes
The boston_dm service has this, the new DM services (#156) don't.

ivalaginja:
Fixes #262.

Since we broke up the way voltages and DM surface is computed in the new services, I was unsure where to add the trace interval previously labelled as "compute voltages":
https://github.com/spacetelescope/catkit2/blob/67833ed37cd2fc02e26b03b109df344895b5aa04/catkit2/services/bmc_dm/bmc_dm.py#L100

I have put the others in their respective new place:
- "send surface"
- "update dm surface"
- "compute total surface"
- "send data"

Edit: The "compute voltages" trace was added as the PR progressed.

### DIFF
--- a/catkit2/base_services/bmc_deformable_mirror.py
+++ b/catkit2/base_services/bmc_deformable_mirror.py
@@ -54,9 +54,13 @@ class BmcDeformableMirror(DeformableMirrorService):
             # Send voltages to the device.
             self.send_to_device()
 
+            with trace_interval('compute voltage'):
+                discretized_surface = self.discretized_surface
+                discretized_voltages = self.discretized_voltages
+
             # Submit discretized surface and voltages to data streams.
-            self.total_surface.submit_data(self.discretized_surface)
-            self.total_voltage.submit_data(self.discretized_voltages)
+            self.total_surface.submit_data(discretized_surface)
+            self.total_voltage.submit_data(discretized_voltages)
 
     def send_to_device(self):
         '''Send the surface to the simulated/real hardware.
--- a/catkit2/base_services/bmc_deformable_mirror.py
+++ b/catkit2/base_services/bmc_deformable_mirror.py
@@ -1,4 +1,5 @@
 from catkit2.base_services.deformable_mirror import DeformableMirrorService
+from catkit2.testbed.tracing import trace_interval
 
 import numpy as np
 from astropy.io import fits
@@ -47,14 +48,15 @@ class BmcDeformableMirror(DeformableMirrorService):
         surface : ndarray
             The requested surface of the DM(s).
         '''
-        self.surface = surface
+        with trace_interval('send surface'):
+            self.surface = surface
 
-        # Send voltages to the device.
-        self.send_to_device()
+            # Send voltages to the device.
+            self.send_to_device()
 
-        # Submit discretized surface and voltages to data streams.
-        self.total_surface.submit_data(self.discretized_surface)
-        self.total_voltage.submit_data(self.discretized_voltages)
+            # Submit discretized surface and voltages to data streams.
+            self.total_surface.submit_data(self.discretized_surface)
+            self.total_voltage.submit_data(self.discretized_voltages)
 
     def send_to_device(self):
         '''Send the surface to the simulated/real hardware.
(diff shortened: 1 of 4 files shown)


catkit2/base_services/bmc_deformable_mirror.py
```diff
@@ -47,14 +48,15 @@ def send_surface(self, surface):
...
             The requested surface of the DM(s).
         '''
-        self.surface = surface
+        with trace_interval('send surface'):
+            self.surface = surface
 
-        # Send voltages to the device.
-        self.send_to_device()
+            # Send voltages to the device.
+            self.send_to_device()
```
raphaelpclt, line 56: I think I would squeeze here the "compute voltage" trace interval, by breaking it down into:
```
with trace_interval('compute voltage'):
    discretized_surface = self.discretized_surface
    discretized_voltage = self.discretized_voltage
```

and only then submit to the datastreams

ehpor: seconded

ivalaginja: @raphaelpclt, very sorry, but I don't understand what you mean. What's the code you posted?

raphaelpclt: Oh in the description of the PR, you mention you don't know where to port the `compute voltage` trace. I would put it here. With more context:

```
with trace_interval('send surface'):
    self.surface = surface

    # Send voltages to the device.
    self.send_to_device()

with trace_interval('compute voltage'):
    discretized_surface = self.discretized_surface
    discretized_voltage = self.discretized_voltage

self.total_surface.submit_data(discretized_surface)
self.total_voltage.submit_data(discretized_voltages)
```

ivalaginja: Ah ok. I was just confused because it looked like you wanted me to add code where it didn't exist before. Now I see that's indeed what you meant. Thanks for the clarification!

raphaelpclt: Does it make sense?

ivalaginja: This means the data stream submission is not included in the tracing anymore, is that what we want? Or should I indent it in such a way that both the "compute voltage" trace as well as the data stream submission is inside the "send surface" trace indentation?

raphaelpclt: I wouldn't include the datastream submission in it. That's how it's done in https://github.com/spacetelescope/catkit2/blob/67833ed37cd2fc02e26b03b109df344895b5aa04/catkit2/services/bmc_dm/bmc_dm.py#L100.

ivalaginja: Oh wow, I was reading that wrong.

ivalaginja: But the old one did have "compute voltages" inside of "send surface" - so should I indent it one level compared to what you posted above?

raphaelpclt: Oh yes good catch, you're right

ivalaginja: Ok all is clear now. Thank you for your patience 🙏

ivalaginja: However, will flake8 yell at me for a unused variables? 😅

ivalaginja: I guess I could to the following instead?

```python
            with trace_interval('compute voltage'):
                self.discretized_surface
                self.discretized_voltage
```

raphaelpclt: Euh no no you need to submit these variables to the datastream and not the attributes. (ie the `discretized_voltages` and not `self.discretized_voltages`) otherwise you compute them twice.  That's the whole point of breaking down the code

ivalaginja: My bad...

raphaelpclt (approved):
Ok

raphaelpclt:
From a test on HiCAT:
Screenshot 2025-02-05 at 5 52 43 PM

PR #283, opened 2025-02-03
Add trace interval 'processing frame' to Hamamatsu camera

ivalaginja:
Like done for the FLIR and ZWO cameras.

The AV camera service uses a frame handler and I had no clue how to deal with that, otherwise I would have added it there too.

Needs testing.


ehpor (approved):
Seems like a trivial change that does not need a test.

PR #286, opened 2025-02-11
MNT: Replace ubuntu-20.04 with ubuntu-22.04

pllim:
As recommended by actions/runner-images#11101 , this PR replaces the deprecated ubuntu-20.04 runner with a newer version. This also replaces RTD workflow, if applicable, to future-proof the build in case it follows suit.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

[:ghost:](https://github.com/pllim/playpen/issues/63)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: ".github/workflows" # Location of package manifests
+    schedule:
+      interval: "monthly"
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -60,7 +60,7 @@ jobs:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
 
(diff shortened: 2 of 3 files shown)


ivalaginja:
The hardware that this repo is primarily made for is limited to Python 3.7.6, and the CI is supposed to reflect that, so this won't work. I will be so free to close this PR without merging for now.

pllim:
No worries but please be aware that GitHub is actually removing ubuntu-20.04. This is not under our control. That means your workflows will soon break. FYI.

ivalaginja:
@pllim thanks for the heads up. Do you have any more precise info about what "soon" means? Thanks!

pllim:
This is all explained in https://github.com/actions/runner-images/issues/11101 linked above.

*We will soon start the deprecation process for Ubuntu 20.04. While the image is being deprecated, you may experience longer queue times during peak usage hours. Deprecation will begin on 2025-02-01 and the image will be fully unsupported by 2025-04-01.*

ivalaginja:
Thank you very much @pllim!

@RemiSoummer fyi since we talked about the versioning mess *very* recently...

pllim:
I am curious why your code is tied into Python 3.7 only. It is past end-of-life and is now exposed to security risks because it is no longer being patched. You are also tied to many old versions of other things because they have also dropped Python 3.7 support in the past few years. If you wish to consult on how to upgrade, I am sure STScI devs will be happy to advise. Good luck.

ivalaginja:
Like said above, it's a single piece of hardware that is unsupported beyond a certain Python version. Rest assured that the team is fully aware of the constraints you list, we've been through dependency hell before, but we can't do much when the vendor is behind on their API releases.

PR #287, opened 2025-02-12
Add PCIe device id choice to DM startup

ivalaginja:
When more than one PCIe card for separate DMs is used on a single machine, the DM serial number is not enough to identify the connection. The device ID of the corresponding PCIe card needs to be provided to the BMC device prior to opening it, as documented in the `bmc` module.

This PR adds this functionality, with a default to ignore itself if no PCIe device ID is provided, in which case the controller automatically connects to the only card available.

Important:
https://github.com/spacetelescope/catkit2/pull/287#issuecomment-2653934895

### DIFF
--- a/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
+++ b/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
@@ -21,6 +21,7 @@ class BmcDeformableMirrorHardware(BmcDeformableMirror):
     def __init__(self, service_type='bmc_deformable_mirror_hardware'):
         super().__init__(service_type)
 
+        self.device_id = self.config.get('device_id')
         self.device_command_index = self.config.get('device_command_index', 0)
         self.enable_high_resolution = self.config.get('enable_high_resolution', False)
 
@@ -31,6 +32,7 @@ class BmcDeformableMirrorHardware(BmcDeformableMirror):
 
     def open(self):
         self.device = bmc.BmcDm()
+        status1 = self.device.set_device_id(self.device_id)
         status = self.device.open_dm(self.serial_number)
 
         if status != bmc.NO_ERR:
--- a/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
+++ b/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
@@ -21,7 +21,6 @@ class BmcDeformableMirrorHardware(BmcDeformableMirror):
     def __init__(self, service_type='bmc_deformable_mirror_hardware'):
         super().__init__(service_type)
 
-        self.device_id = self.config.get('device_id')
         self.device_command_index = self.config.get('device_command_index', 0)
         self.enable_high_resolution = self.config.get('enable_high_resolution', False)
 
--- a/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
+++ b/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
@@ -31,9 +31,12 @@ class BmcDeformableMirrorHardware(BmcDeformableMirror):
 
     def open(self):
         self.device = bmc.BmcDm()
-        status1 = self.device.set_device_id(self.device_id)
-        status = self.device.open_dm(self.serial_number)
 
+        status1 = self.device.set_device_id(self.device_id)
+        if status1 != None:
+            pass
+
+        status = self.device.open_dm(self.serial_number)
         if status != bmc.NO_ERR:
             raise RuntimeError(f'Failed to connect: {self.dm.error_string(status)}.')
 
--- a/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
+++ b/catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
@@ -32,13 +32,13 @@ class BmcDeformableMirrorHardware(BmcDeformableMirror):
     def open(self):
         self.device = bmc.BmcDm()
 
-        status1 = self.device.set_device_id(self.device_id)
-        if status1 != None:
-            pass
+        pcie_status = self.device.set_device_id(self.device_id)
+        if pcie_status != bmc.NO_ERR:
+            raise RuntimeError(f'Failed to connect: {self.dm.error_string(pcie_status)}.')
 
-        status = self.device.open_dm(self.serial_number)
-        if status != bmc.NO_ERR:
-            raise RuntimeError(f'Failed to connect: {self.dm.error_string(status)}.')
+        dm_status = self.device.open_dm(self.serial_number)
+        if dm_status != bmc.NO_ERR:
+            raise RuntimeError(f'Failed to connect: {self.dm.error_string(dm_status)}.')
 
         self.device.enable_high_res(enable=self.enable_high_resolution)
         self.device_command_length = self.device.num_actuators()
--- a/catkit2/base_services/bmc_deformable_mirror.py
+++ b/catkit2/base_services/bmc_deformable_mirror.py
@@ -10,6 +10,7 @@ class BmcDeformableMirror(DeformableMirrorService):
         super().__init__(service_type)
 
         self.serial_number = self.config['serial_number']
+        self.device_id = self.config.get('device_id', 0)
         self.flat_map_fname = self.config['flat_map_fname']
         self.gain_map_fname = self.config['gain_map_fname']
         self.max_volts = self.config['max_volts']


ivalaginja:
We tested this in the lab today on two DMs connected through two distinct PCIe cards and it works as expected. I would just like to run it on the latest commit one more time to definitively put a green light on this.

ivalaginja:
Tested on latest commit on hardware and it works.

ivalaginja:
Not entirely sure what that failing MacOS test is though...

raphaelpclt (approved):
Looks good

ivalaginja:
As a side note, the PCIe numbering changes when rebooting the machine, so this will pose a problem... Doesn't impact the functionality introduced here though.

catkit2/services/bmc_deformable_mirror_hardware/bmc_deformable_mirror_hardware.py
```diff
@@ -31,10 +31,14 @@ def __init__(self, service_type='bmc_deformable_mirror_hardware'):
 
     def open(self):
         self.device = bmc.BmcDm()
-        status = self.device.open_dm(self.serial_number)
 
-        if status != bmc.NO_ERR:
-            raise RuntimeError(f'Failed to connect: {self.dm.error_string(status)}.')
+        pcie_status = self.device.set_device_id(self.device_id)
+        if pcie_status != bmc.NO_ERR:
+            raise RuntimeError(f'Failed to connect: {self.dm.error_string(pcie_status)}.')
```
steigersg, line 37: @ivalaginja @raphaelpclt - On HiCAT I'm getting an AtrributeError: 'BmcDeformableMirrorHardware' object has no attribute 'dm'. I noticed in this PR there is a commit that states "Not to be merged in catkit2". Was this committed in error? or is there something else going on that needs to be fixed independently?

ivalaginja: That commit messages should have just stated "WIP" instead :P
This looks like something unrelated - HiCAT doesn't have a device ID set for the DM so it should just default to 0. If the error is about the attribute `dm`, I do think it's something else.

steigersg: Sounds good, as I was debugging that message obviously set off a red flag but I'll keep digging - thanks!

ivalaginja: Ooh wait. `self.dm` looks like  it should be `self.device` instead in the first place. Is this an error from the DM base class refactor? It looks like it.
In any case, it does mean that you are getting an error when trying to set the PCIe device id, so it definitely *is* related to what's been introduced in this PR.

ivalaginja: Git blame says it's me who introduced this bug 😬 So that needs fixing (`self.dm` --> `self.device`).
This won't tell you why you get a PCIe device error though. Let me get back to you on how to check this shortly.

ivalaginja: Depending on what you're exactly working on, it would be worth doing the above-mentioned fix and then let the code actually raise the error, so that we see what it says. My guess is that the current default devId coded into the bmc hardware service is not the automatic devId used by HiCAT. This is easy to fix but we should check what exactly is happening.

ivalaginja: Oh and also start up the DM service in its own terminal, this should give you some services logs like "Using automatic devId: x", that would be useful (and some lines before and after this one).

steigersg: After updating `self.dm` --> `self.device`

Error when trying to start the service in its own window

Screenshot 2025-03-31 at 11 14 29 AM

Error in server window when trying to apply DM flatmaps/start the DM service 

Screenshot 2025-03-31 at 11 17 34 AM

steigersg: Also maybe worth noting that Rémi and I on Friday did a full reinstall of all 3 repos so everything should be up to date

ivalaginja: This seems to come from the --id argument, I believe the service name for the DMs on HiCAT is `boston_dm`?

steigersg: should I be running the base service here (`base_services/bmc_deformable_mirror.py`) or the hardware service (`bmc_defromable_mirror_hardware/bmc_deformable_mirror_hardware.py`)?

ivalaginja: If you're in hardware, definitely the hardware sevice, which you are, as you're calling the right .py file. But in the last example it seems like you tried to link it to a service ID that doesn't exist in the configi file (bmc_deformable_mirror_hardware). The configuration for the Bostons is given by the section "boston_dm" (maybe with an s at the end, I don't remember).

Unless you added an ID with that name of course! Then it will be something else.

steigersg: Apologies, but I'm not used to using the service debugging - that "debug" service needs to be started before starting the server right? I'm able to start the bmc hardware service in the separate window now, but it gets killed when I start the server because the tested is "unsafe" (just think the sensors need a second to get readings as is often seen in the GUI) and then I am unable to restart it and get the same `RuntimeError: Failed to connect: Bad argument...`. 

This is irregardless of if I give the "debug" service a random port for the service or if I match the service port in the debug window to one of the DM ports the server opens.

ivalaginja: No worries, I think we didn't touch those in quite a while...

In all cases, you need to start the testbed server first. Since it's the broker between the different parts of the system, nothing can run without it. Only after starting the server can a service register with the server during its own startup, no matter whether started through the GUI, an experiment or in its own debug window. Indeed, the concrete service port number doesn't matter, as long as you're not giving it an already occupied port, which would be unlucky coincidence.

I think you're on the right path here: start the testbed server first, then in a separate window, start the DM service with a random service port number. Let me know if that works!

steigersg: Great, that was my understanding as well so happy I was on the right track - still get the following however 

Screenshot 2025-03-31 at 12 02 36 PM

ivalaginja: Since I have limited understanding of the port attributions, could you try to give it a service port number with only 4 characters (I doubt it's that, but doesn't hurt to try...)

steigersg: just did - identical message

ivalaginja: Hm, I am starting to run out of things to try, unfortunately.

steigersg: that's alright - I appreciate your help! I'll keep looking into it and will comment the solution here/ Make a PR when I figure it out

steigersg: So line 35 is the issue (though I'm still not quite sure why since I need to look into what `set_device_id` does), but if I change 

```
pcie_status = self.device.set_device_id(self.device_id)
if pcie_status != bmc.NO_ERR:
    raise RuntimeError(f'Failed to connect: {self.dm.error_string(pcie_status)}.')
```
to 
```
if self.device_id:
  pcie_status = self.device.set_device_id(self.device_id)
  if pcie_status != bmc.NO_ERR:
      raise RuntimeError(f'Failed to connect: {self.dm.error_string(pcie_status)}.')
```
I get no errors

ivalaginja: Hmmm how is it possible this makes a difference? `self.device_id` is always something, since it initializes with 0 if it doesn't get a value from the config, so the if should always be entered. Or am I seeing something wrong?

steigersg: If its 0 Python will treat it as a boolean in an `if` statement 

Screenshot 2025-03-31 at 2 30 35 PM

ivalaginja: Lol good ol' Python. I forgot.

Could I maybe ask you to send me the DM service log in the case where it opens ok? I think there should be some deviceID information in there that could help us debug, even when it opens fine and doesn't run through `set_device_id()`.

PR #288, opened 2025-02-15
MNT: Replace ubuntu-20.04 with ubuntu-22.04 (try 2)

pllim:
I just want to see if plainly upgrading ubuntu without touching python would work. If it does, then you have a bit more time to deal with your hardware woes.

Also the workflow_dispatch syntax was wrong, so I fixed it too.

Another try of #286

[👻](https://github.com/pllim/playpen/issues/63)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -60,7 +60,7 @@ jobs:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     needs: build_docs
     steps:
       - name: Deploy to GitHub Pages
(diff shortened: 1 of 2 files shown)


pllim:
Hmm `ready_for_review` flag to rerun CI is unnecessary but okay.

ivalaginja:
@ehpor do you remember whether the FLIR can take py3.7.16? I feel like you were the one down the rabbit hole identifying compatible minor versions.

ehpor:
I see no reason why it couldn't. The bugfix version pin was something leftover from the catkit1 days. I feel we can be more free than that.

pllim:
The new minor version was copied from your other jobs that do pass, so I don't see why it would be a problem. 🤷‍♀️

ehpor (approved):
LGTM. These are docs and linting, so they'll work just fine with any Python version. Not critical.

ivalaginja:
Thanks for figuring this out for us @pllim!

PR #289, opened 2025-02-21
Add shareable base class

ehpor:
This PR adds a base class for all objects that are shared between processes. A shareable object has a shared state and a local state. The shared state is assumed to be in shared memory and is accessed by the shareable object using a pointer. The local state is internal to the shareable object. Shareable objects have `Create()` and `Open()` methods that take a pointer to that shared state to initialize or open. Note: this base class provides no methods for actually sharing this information between processes. That's the job of a `SharedMemory` or other similar classes. It just formalizes the size and open methods, and provides an enum for all shareable types.

This PR also adds an `Event` class that combines all other Event implementations into one single class. An Event allows you to wait using multiple methods, choosable at runtime. This will be useful for testing comparing synchronization methods in practice rather than only during tests (like I've been doing so far).

The data stream is also changed to use _two_ shared memory blocks per stream. The first one is for header information, communicating where each data frame is in the buffer. The second is the buffer where the actual data is stored. This allows us to support GPU shared memory in the future (since the header information should not be stored in GPU shared memory but rather on CPU shared memory. Therefore, two separate shared memory buffers is required.

**Risk to existing testbeds:** no risk. Everything should be refactor only, all modifications are C++ only with no side-effects leaking into Python code.


ehpor:
@raphaelpclt @ivalaginja This is ready for your review now. I'll do some more testing of edge cases, but I do not see this changing anything significantly other than 1-2 lines here and there, if any.

ivalaginja:
Hey @ehpor happy to look at this, but since it's all in the C++ part of the package, do you have any pointers what I should be paying particular attention to?

@raphaelpclt is on vacation for the next two weeks.

ehpor:
@ivalaginja The main class to start with is the `Shareable` class, and the simplest implementation is the `SharedMemory` class, even though it's a little of an odd duck in the pack. Another good one is the `Event` class, which is also simple.

We can also schedule a call tomorrow morning sometime after 6pm Paris time to go over some of the details together, but here is a short description to make the changes easier to understand.

The main goal for this PR is to have a single mechanism to share objects between processes. In the current code base, there are getting to be more and more objects that need to be shared in shared memory, which created multiple ways to construct them and how to hold shared state, including multiple nomenclatures. In this new system, all objects that are supposed to be shared between processes, all derive from `Shareable` and have a single mechanism for creation and shared state. In short, the new mechanism uses:

* A `Create()` function that creates a new object from scratch, with some of the information contained in a memory block outside of that object's memory. 
* An `Open()` function takes in a pointer to that external piece of memory and reconstructs its own view of that object. 

Since memory can be shared between processes, this allows us to do:

1. Process A creates a piece of shared memory under a specific name.
2. Process A creates object X using a portion of that shared memory for its shared state.
3. Process B opens the piece of shared memory created by process A.
4. Process B opens object X using the shared state in the shared memory.

After this, both process A & B have access to the same object. That means that the processes can communicate with each other via object X and its shared state.

Currently, I converted over:
1. The `DataStream`.
2. The `Event`.
3. The `FreeListAllocator`.
4. The `PoolAllocator`.
5. The `SharedMemory` itself.
6. The `HashMap`.

In the future, there will also be a `MessageBroker` and maybe other helper classes.

raphaelpclt (approved):
Tested on HiCAT with @alexisyslau, and went through the code with @ehpor

PR #290, opened 2025-02-24
Extend service to NKT fianium

pbaudoz:
> [!NOTE]  
> Superseded by #373

This PR aims to setup a service for Fianium + Varia config.

### Todo:
- [x] Adjust simulated service to follow changes in hardware service.
- [x] Update service docs for the NKT.


ehpor:
I feel like combining the Fanium with the current `nkt_superk` service should be doable. It'd just require changing an enum and setting the right values when creating the methods on the class. Right now, creating the functions is done as

https://github.com/spacetelescope/catkit2/blob/e3f59121abfc37381a71888604cba6e2e9d8a727/catkit2/services/nkt_superk/nkt_superk.py#L235-L237

etc...

But you can do the same in the `__init__()`. Something like:

```python
def __init__(self, ...):
    ... # Other setup code.
    
    connected_device = self.config['connected_device'].lower()
    if connected_device == 'evo':
        device = Evo
    elif connected_device == 'fanium':
        device = Fanium
    else:
        raise NotImplementedError(f'Connected device {connected_device} not implemented.')
    
    # Common functions between Evo and Fanium.
    self.get_emission = read_register(registerReadU8, device.REG_EMISSION, ratio=0.5)
    self.get_emission = read_register(registerReadU8, device.REG_EMISSION, ratio=0.5)
    
    if device is Evo:
        self.get_base_temperature = read_register(registerReadS16, Evo.REG_BASE_TEMPERATURE, ratio=0.1)
    
    if device is Fanium:
        ... # Fanium specific functions.
```

Does that work for you?

catkit2/services/nkt_superk/nkt_superk.py
```diff
@@ -46,6 +46,36 @@ class Evo(Enum):
     REG_MAC_ADDRESS = 0xB3
 
 
+class Fianium(Enum):
+    DEVICE_ID = 15
+
+    #REG_BASE_TEMPERATURE = 0x17
+    #REG_SUPPLY_VOLTAGE = 0x1D
+    #REG_EXTERNAL_CONTROL_INPUT = 0x94
```
ivalaginja: If you have double-checked that the Fianium doesn't have any of these, delete them all. Together with the "added" comments, since that is obvious from the code.

catkit2/services/nkt_superk/nkt_superk.py
```diff
@@ -138,11 +235,16 @@ def open(self):
...
             'emission': self.monitor_func(self.emission, self.set_emission),
             'power_setpoint': self.monitor_func(self.power_setpoint, self.set_power_setpoint),
-            'current_setpoint': self.monitor_func(self.current_setpoint, self.set_current_setpoint),
             'varia_status': self.update_func(self.update_varia_status),
-            'evo_status': self.update_func(self.update_evo_status)
         }
 
+        if self.device is Evo:
+            funcs['current_setpoint'] = self.monitor_func(self.current_setpoint, self.set_current_setpoint)
+            funcs['evo_status'] = self.update_func(self.update_evo_status)
```
ivalaginja, line 224: I am pretty sure we'll need an update function for the Fianium as well. For that you will need to check what you can read from it with some sort of "get" functions I suppose, similar to the Evo.

catkit2/services/nkt_superk_fianium/nkt_superk_fianium.py
ivalaginja, line 1: Don't forget to delete this file!

ivalaginja (changes requested):
Looks great so far! Left some comments for further steps. We can go through how to work through failed CI tests and how to adapt the simulated service later.

catkit2/services/nkt_superk/nkt_superk.py
ivalaginja: Do we not need an update function for the Fianium in the end?

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -7,18 +7,22 @@
 class NktSuperkProxy(ServiceProxy):
     @property
     def center_wavelength(self):
+        print('Getting central wavelength...')
```
ivalaginja: These prints everywhere look like used for debugging, they should be deleted again. If you want to keep them they'd need to go to the log.

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -68,3 +72,60 @@ def sleep_time_per_nm(self):
...
+                print('Could not getting pulse picker ratio')
+
+    @pulse_picker_value.setter
+    def pulse_picker_value(self, pulse_picker_ratio_to_set):
+        print('Setting pulse_picker_ratio value...')
+        try:
+            self.pulse_picker_ratio.submit_data(np.array([pulse_picker_ratio_to_set], dtype='uint16'))
+        except RuntimeError:
+                # The frame wasn't available anymore because we were waiting too long.
+                print('Could not setting pulse picker ratio')
```
ivalaginja: Only the Fianium has a pulse picker ratio, so this won't work as a general property.

Since it only submits to or reads something from a data stream, you could just delete it altogether. On the testbed side, we can then simply submit to the data stream or read from it instead of adding the extra layer of a proxy function. This would be the easy solution. Anything else would be more complicated.

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -68,3 +72,60 @@ def sleep_time_per_nm(self):
...
         return self.config['base_sleep_time']
+
+    @property
+    def emission_value(self):
+        return self.emission.get()[0]
+
+    @emission_value.setter
+    def emission_value(self, emission_to_set):
+        print('Setting emission value...')
+        self.emission.submit_data(np.array([emission_to_set], dtype='uint8'))
```
ivalaginja: I think you can delete this property - I'll leave you a note in the THD2 PR how to access the emission data stream in the general light source proxy instead.

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -68,3 +72,60 @@ def sleep_time_per_nm(self):
...
+        else:
+            print('NKT already ON')
+
+    def turn_off(self):
+        emission_value = self.emission_value
+        if emission_value != [0]:
+            self.emission.submit_data(np.array([0], dtype='uint8'))
+            print('NKT turned OFF')
+        else:
+            print('NKT already OFF')
```
ivalaginja: All of our light source commands in the GUI and experiments are set up to send values to the `emission` data stream directly, or to read from it. Both of these operations can be done in a single line. I thus think that these two extra methods are not needed and just add an extra layer of code that is not super useful. I would recommend to delete them, and like for the emission, I will show you over on THD2 how to achieve the same effect without them.

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -68,3 +72,60 @@ def sleep_time_per_nm(self):
...
+
+    @property
+    def power(self):
+        print('Getting power value...')
+        return self.power_setpoint.get()[0]
+
+    @power.setter
+    def power(self, power_to_set):
+        print('Setting power value...')
+        self.power_setpoint.submit_data(np.array([power_to_set], dtype='float32'))
```
ivalaginja: I believe here too that we can probably ditch this property, but I will want to check with you.

The reasoning behind all of my comments behind deleting proxy properties is that proxy functions and properties wrap service functionality so that we don't have to write the code for that functionality repeatedly. One example for this is the `set_spectrum()` function in this very proxy for the NKT. You will see that it does a bunch of different things to actually set the spectrum, and you definitely don't want to repeat that code each time you are setting the spectrum. When the code to achieve a hardware effect is only writing to a data stream though, adding the additional layer of a proxy function is not super necessary and just adds empty layers. Remember that going through a proxy is not required! It just adds convenience.

Let's chat over on the THD2 PR about how to adjust the code there so that you don't need the ones I flagged here anymore.

ivalaginja (commented):
I focused my review on the proxy since that is tightly connected to usage on THD2. The service changes look ok to me but I would like @ehpor's opinion on this, and ideally a test on HiCAT to make sure we're not braking their usage.

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -68,3 +72,60 @@ def sleep_time_per_nm(self):
...
+        emission_value = self.emission_value
+        if emission_value != [0]:
+            self.emission.submit_data(np.array([0], dtype='uint8'))
+            print('NKT turned OFF')
+        else:
+            print('NKT already OFF')
+
+
+    @property
+    def power(self):
```
ivalaginja: Also, just for me to understand: would we ever have to set the power directly?

catkit2/testbed/proxies/nkt_superk.py
```diff
@@ -68,3 +72,60 @@ def sleep_time_per_nm(self):
...
+        return self.power_setpoint.get()[0]
+
+    @power.setter
+    def power(self, power_to_set):
+        print('Setting power value...')
+        self.power_setpoint.submit_data(np.array([power_to_set], dtype='float32'))
+
+
+    @property
+    def pulse_picker_value(self):
```
ivalaginja: Same question here, is there a use case where we would want to set the pulse picker ratio directly, apart from reading it from the config file upon starting of the service?

pbaudoz: Yes, each time we record an off-axis image we want to be able to increase the pulse picker value and when coming back to on-axis decrease it

catkit2/testbed/proxies/nkt_superk.py
pbaudoz: Not sure since variation of the power changes the spectrum.

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Ok. What does this ratio change? Why does it need to be different for direct PSFs?

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Would it make sense for now to pick one value and stick with it?

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Btw this is exactly why the `get_flux_mode()` and `set_flux_mode()` functions exist in the light source proxy (on thd-controls), so it will be easy to put this in there with the data stream directly. No need for this property here.

catkit2/services/nkt_superk/nkt_superk.py
```diff
@@ -165,7 +244,7 @@ def main(self):
 
     def close(self):
         # Turn off the source
-        self.set_emission(0)
+        self.set_emission(self, 0)
```
ehpor: I'm fairly certain that this doesn't work like that. The `self.set_emission` part already sets `self` as the first argument. So passing it explicitly is wrong. The same is true for the other calls further down.

ehpor (commented):
I agree with Iva's other comments.

catkit2/services/nkt_superk/nkt_superk.py
axelpotier-astro: Indeed! Changed it in commit 261091d13752abb5b04329f9b8a842cd30c4584f

catkit2/testbed/proxies/nkt_superk.py
axelpotier-astro: I kept a function in thd-control proxy, in case we need it at some point

catkit2/testbed/proxies/nkt_superk.py
axelpotier-astro: I put them in the thd2 PR

catkit2/testbed/proxies/nkt_superk.py
axelpotier-astro: I believe the turn on function is important because a 3 (and not a 1!) must be sent to the datastream. I put these functions in thd-control

catkit2/testbed/proxies/nkt_superk.py
axelpotier-astro: I do not know how to put them in the log

catkit2/services/nkt_superk/nkt_superk.py
axelpotier-astro: I do not know what an update function does. It has not been required so far

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Then just delete the prints for now.

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: This still needs to be deleted.

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Ok, then just make sure to delete it here.

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Delete here please.

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: ok then delete here please 😬

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Talked to @pbaudoz and I understand now a bit better - this still needs to be handled on the THD2 side and deleted here though.

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: And again, please actually delete.

ivalaginja (commented):
I see the changes on thd-controls, but I think you still need to delete the marked stuff here?

catkit2/testbed/proxies/nkt_superk.py
ivalaginja: Oh, then I must have misunderstood something. Is this the same for the one on HiCAT @ehpor? Even more reason to have this in the custom proxy on thd-controls.

ivalaginja:
@axelpotier-astro this is not in the way of the hardware tests, but we'll also have to address the two remaining TODO items from the PR description: updating the simulated service and adjusting the decoumentation for this service.

Updating the simulated NKT service (https://github.com/spacetelescope/catkit2/blob/develop/catkit2/services/nkt_superk_sim/nkt_superk_sim.py) should reflect the changes introduced here. In particular, it should follow the same logic of choosing between the Evo and the Fianium and create the according data streams. Any communication to hardware does obviously not happen.

We probably won't be able to fully reflect the power changes due to the pulse picker ratio right now, but we should at least make sure that the `init()` and `open()` methods are the same.

The doc changes should be obvious, you can find the according rst file in docs/services/service type.

catkit2/services/nkt_superk/nkt_superk.py
ivalaginja: The update function of the Evo updates the data stream values for parameters that give the general status of the Evo (base temperature, supply voltage and external control input). Since there is nothing equivalent defined for the Fianium, we indeed don't need an update function for it, at least for now.

ivalaginja (commented):
Fine by me, but would like to have @ehpor look over it one more time.

THD2-team:
@ehpor we are in the lab, @axelpotier-astro , @pbaudoz and @ivalaginja and we don't understand why we can't make the service work. This has been an issue from the very beginning which is why we had added the `self`s where they don't belong. Do you have any idea about what is happening? Below the console output when we try starting the NKT on this branch in a separate terminal window. Note how the service status switches to "RUNNING" as if everything was fine - it just doesn't seem to be able to spawn the threads.

```bash

thd2@thd2-controller MINGW64 ~
$ conda activate catkit2
(catkit2)
thd2@thd2-controller MINGW64 ~
$ cd repos/catkit2/catkit2/services/nkt_superk
(catkit2)
thd2@thd2-controller MINGW64 ~/repos/catkit2/catkit2/services/nkt_superk (nkt_fianium)
$ ls
nkt_superk.py
(catkit2)
thd2@thd2-controller MINGW64 ~/repos/catkit2/catkit2/services/nkt_superk (nkt_fianium)
$ python nkt_superk.py --id nkt_superk --port 3845 --testbed_port 8768
Loading x64 DLL from: C:\Program Files (x86)\NKT Photonics\SDK\NKTPDLL\x64\NKTPDLL.dll
Function GetSocket in C:\Users\thd2\repos\catkit2\catkit_core\Client.cpp:101
  Debug: Creating new socket.
Function Service in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:54
  Debug: Registering request handlers.
Function Service in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:60
  Info: Intialized service.
Function Run in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:91
  Info: This service can operate in unsafe conditions.
Function Run in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:104
  Info: Opening service.
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "pulse_picker_ratio".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "emission".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "power_setpoint".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "monitor_input".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "nd_setpoint".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "swp_setpoint".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "lwp_setpoint".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "nd_filter_moving".
Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "swp_filter_moving".


Function MakeDataStream in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:512
  Debug: Making data stream "lwp_filter_moving".
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
Exception in thread Thread-2:
Traceback (most recent call last):
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
Exception in thread Thread-3:
    self.run()
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
Traceback (most recent call last):
Exception in thread Thread-4:
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
Traceback (most recent call last):
Exception in thread Thread-6:
Traceback (most recent call last):
Function Run in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
121
  Info: Service was succesfully opened.
    self.run()
Function Run in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:14Exception in thread Thread-7:
2
  elf._target(*self._args, **self._kwargs)
3    self.run()
4  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
xception in thread Thread-5: properties.
3  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 305, in func
9m
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
    self.run()
Function Run  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
 in C:\Users\thd2\repos\catkit2\catkit_core\Service.cpp:18    setter(frame.data[0])
5
  Function   File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
Ruraceback (most recent call last):catypeError: setter() missing 1 required positional argument: 'value'
34m    self.run()
39m     nfo: Starting service main function.ile "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
39m49m
    self._target(*self._args, **self._kwargs)

Traceback (most recent call last):
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 305, in func
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 980, in _bootstrap_inner
    self._target(*self._args, **self._kwargs)
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 305, in func
    self._target(*self._args, **self._kwargs)
    self.run()
    setter(frame.data[0])
    self._target(*self._args, **self._kwargs)
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
    self.run()
    setter(frame.data[0])
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 305, in func
  File "C:\Users\thd2\miniconda3\envs\catkit2\lib\threading.py", line 917, in run
TypeError: setter() missing 1 required positional argument: 'value'
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 328, in func
TypeError: setter() missing 1 required positional argument: 'value'
    setter(frame.data[0])
TypeError: setter() missing 1 required positional argument: 'value'
    updater()
    self._target(*self._args, **self._kwargs)
    self._target(*self._args, **self._kwargs)
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 305, in func
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 319, in func
  File "C:\Users\thd2\repos\catkit2\catkit2\services\nkt_superk\nkt_superk.py", line 281, in update_varia_status
    setter(frame.data[0])
TypeError: setter() missing 1 required positional argument: 'value'
    setter(frame.data[0])
    status = self.get_varia_status_bits()
TypeError: setter() missing 1 required positional argument: 'value'
TypeError: getter() missing 1 required positional argument: 'self'
```

ehpor:
You are setting getters and setters as attributes on the instance, not on the class, like before. The self gets added when they are class attributes, not when they are instance attributes.

THD2-team:
> You are setting getters and setters as attributes on the instance, not on the class, like before. The self gets added when they are class attributes, not when they are instance attributes.

...

Understood. We'll just need help in fixing it. How can we keep them attributes on the class while being able to differentiate between the Evo and Fianium?

THD2-team:
Thank you @ehpor. Ready for another review.

Tested on hardware and works.

ivalaginja:
Follow-up in #372

ivalaginja:
Superseded by https://github.com/spacetelescope/catkit2/pull/373.

PR #294, opened 2025-03-11
Add line to installation.rst describing how to increase ulimit

steigersg:
Adds a few lines of docs describing how to increase the ulimit on MacOS upon opening a terminal so you do not have to do it manually. This solves the "too many open files" error encountered upon running services with many processes.


RemiSoummer (approved):
This is good

Issue #293, opened 2025-03-11
Installation on Ubuntu 24.04


rgalicher:
I followed the instruction in the README.md

Everything works up to :
conda env create --file environment.yml
conda activate catkit2

When I launch
pip install -e .

There is an error. See the screen shot below

![Image](https://github.com/user-attachments/assets/aa6e5fb3-89e7-4cb3-956d-96e78c589dca)

raphaelpclt:
Do you have more output by any chance? There are currently 2 different errors we're trying to solve, and the difference is usually further down in the log file.

rgalicher:
Here, is the complete output

![Image](https://github.com/user-attachments/assets/95d04054-b4af-4b72-9469-af5b6b2496ee)

![Image](https://github.com/user-attachments/assets/d59e3212-3f57-4340-8b89-cb0e97104c4b)

![Image](https://github.com/user-attachments/assets/6ff689ec-37d9-4495-a306-0a676173ec6d)

![Image](https://github.com/user-attachments/assets/fd9549d6-9870-467d-baa5-e1f05c2e975d)

![Image](https://github.com/user-attachments/assets/79a5f6ce-e2d3-407f-b02f-1e7465dc7ffb)

![Image](https://github.com/user-attachments/assets/28a96f63-4570-4bd1-876b-dc58b8036141)

raphaelpclt:
Ok we have a quick patch to solve this error - but a deeper solution will be required in the future.

Using an older gcc version solves it:

`sudo apt install gcc-11 g++-11` if you don't have it installed already
`C=/usr/bin/gcc-11 CXX=/usr/bin/g++-11 pip install -e .` (from the catkit2 folder)

rgalicher:
Successfully installed catkit2-0.0.1
WARNING: There was an error checking the latest version of pip.

Thank you

PR #296, opened 2025-03-11
Added missing standard libraries (string, stdint e.t.c.) to .cpp and header files 

alexisyslau:
The current setup works with older compiler versions (see the conversation in #293) but with newer versions, the compilation will fail because some standard libraries were missing in .cpp and header files. 

**Changes I made**
Add standard libraries in .cpp and .h files which will need them. @ehpor please check if there were too many redundancies because I added libs even though it was added to header files.

### DIFF
--- a/catkit_core/LoggingProxy.h
+++ b/catkit_core/LoggingProxy.h
@@ -4,6 +4,7 @@
 #include "Log.h"
 #include "TestbedProxy.h"
 
+#include <string>
 class LoggingProxy
 {
 public:
--- a/catkit_core/LoggingProxy.h
+++ b/catkit_core/LoggingProxy.h
@@ -5,6 +5,7 @@
 #include "TestbedProxy.h"
 
 #include <string>
+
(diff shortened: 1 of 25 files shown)


catkit_core/LoggingProxy.h
```diff
@@ -4,6 +4,7 @@
 #include "Log.h"
 #include "TestbedProxy.h"
 
+#include <string>
```
ehpor, line 7: Add whitespace around the include.

catkit_core/Types.cpp
```diff
@@ -1,4 +1,5 @@
 #include "Types.h"
+#include <string>
```
ehpor, line 3: Add whitespace to separate <> includes and "" includes.

ehpor (changes requested):
Just two whitespace issues, which should take like a minute to fix.

ehpor (approved):
LGTM

PR #291, opened 2025-03-05
Refactor of Shareable states using a Struct specification.

ehpor:
Experimentation on a dynamic version of a C struct, usable for shared state of `Shareable` objects.

This makes heavy use of template metaprogramming to compute the offsets of data elements in a raw buffer. An example usage of this new struct:
```cpp
using NestedStructSpec = StructSpec<
    NamedField<"e", int>,
    NamedField<"f", double, Dynamic>>;

using MyStructSpec = StructSpec<
    NamedField<"a", int>,
    NamedField<"b", double, Dynamic>,
    NamedField<"c", char, 64>,
    NamedField<"d", NestedStructSpec, Dynamic>>;

auto spec = MyStructSpec::Build({
    {"b", 10},
    {"d", 5, {
        {"f", 2}
        }}
    });
```

This PR is meant to be merged into the `feature/shareable` branch before that is merged.

PR #297, opened 2025-03-11
Refactor of Shareable states using a streaming specification.

ehpor:
A streaming alternative to #291, which is cleaner but leads to some duplicated code in each shareable object: the code order for both `Open()` and `Create()`.


alexisyslau (approved):
Should be good to go. 

Looks like all objects using share memory have been migrated to the new structure of the stream. Padding, buffer and alignment are being handled in the header file `StructStream.h` so there should not be any misalignment or not enough buffer. I dont think it would lead to memory issue, maybe more like not that I am aware of 🐒

@ehpor if the compiler is not scream at you, it's good to go!

PR #300, opened 2025-03-13
Fix formatting in installation instructions

ivalaginja:


docs/installation.rst
```diff
@@ -1,20 +1,21 @@
 Installation
 ============
 
-Catkit2 consists of a core C++ library and a Python package that wraps this library. It also requires a number of C++ third party libraries to facilitate JSON encoding and decoding, Python bindings, linear algebra and high-speed communication over sockets. Compilation requires a C++ compiler conforming to the C++20 standard. Catkit2 performs installation of the core library and Python package using CMake and Conda.
+Catkit2 consists of a core C++ library and a Python package that wraps this library. It also requires a number of C++ third party libraries to facilitate JSON encoding and decoding, Python bindings, linear algebra and high-speed communication over sockets. Compilation requires a C++ compiler conforming to the C++17 standard. Catkit2 performs installation of the core library and Python package using CMake and Conda.
```
steigersg, line 4: CATKit2

docs/installation.rst
```diff
@@ -39,25 +40,29 @@ The catkit2 installation requires a pre-installed C++ compiler.
 - On Windows, you can for example install the Visual Studio Compiler, either by installing the Build Tools, or by installing the full IDE with compiler (the community edition is free).
 - On MacOS, nothing should be needed, but some machines require Xcode to be installed. It does install components on first startup, so it is recommended to start XCode on your mac if you have never used it (and accept the license agreement of XCode, which is required).
```
steigersg:
```suggestion
- On MacOS, nothing should be needed, but some machines require Xcode to be installed. It does install components on first startup, so it is recommended to start XCode on your Mac if you have never used it (and accept the license agreement of XCode, which is required).
```

docs/installation.rst
```diff
@@ -39,25 +40,29 @@ The catkit2 installation requires a pre-installed C++ compiler.
...
 - XCode 15 (Sonoma and higher) introduced some changes in the compiler location, which require the following environment variable update:
```
steigersg:
```suggestion
- XCode 15 (Sonoma and higher) introduced some changes in the compiler location, which requires the following environment variable update:
```

docs/installation.rst
steigersg:
```suggestion
CATKit2 consists of a core C++ library and a Python package that wraps this library. It also requires a number of C++ third party libraries to facilitate JSON encoding and decoding, Python bindings, linear algebra and high-speed communication over sockets. Compilation requires a C++ compiler conforming to the C++17 standard. CATKit2 performs installation of the core library and Python package using CMake and Conda.
```

steigersg (changes requested):
Found some small grammatical error and inconsistencies with the capitalization in CATKit2

steigersg (approved):
Looks good to me!

RemiSoummer (approved):
looks good

PR #301, opened 2025-03-19
Improve timestamps to ns-resolution.

ehpor:
Currently, timestamp resolution is 100ns on Windows, and 1000ns on MacOS/Linux, as far as my experience goes (this varies across machine/OS version). This PR uses platform specific code for each OS, that obtains nanosecond resolution timestamps on each OS.

Checking that timestamps are not significantly slower than before:
* MacOS: (MacOS Sonoma, Macbook M2 Pro)
  * Before: 15.4243 ns per timestamp, 1000ns resolution
  * After: 12.7537 ns per timestamp, 1ns resolution
* Windows: (Windows 11, AMD Ryzen 9 5900X 12-Core)
  * Before: 20.9528 ns per timestamp, 100ns resolution
  * After: 23.4281 ns per timestamp, 100ns resolution
* Linux: (Ubuntu Server 20.04, AMD Threadripper 5995WX)
  * Before: 20.7357 ns per timestamp, 1ns resolution
  * After: 21.4882 ns per timestamp, 1ns resolution

### DIFF
--- a/catkit_core/Timing.cpp
+++ b/catkit_core/Timing.cpp
@@ -3,14 +3,96 @@
 #include <sstream>
 #include <iomanip>
 #include <sstream>
+#include <cstdint>
+#include <ctime>
+
+#if defined(__linux)
+	#ifdef CLOCK_MONOTONIC
+		#define CLOCKID CLOCK_MONOTONIC
+	#else
+		#define CLOCKID CLOCK_REALTIME
+	#endif
+#elif defined(__APPLE__)
+	#include <mach/mach_time.h>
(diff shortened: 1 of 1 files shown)


ehpor:
Hmm, after seeing the benchmarks, I'm not sure if we need this or not.

ehpor:
Let's not do this. I'll keep the branch, but we won't merge. The reason is that only on MacOS does this PR improve resolution. And it significantly increases the amount of platform-specific code. Additionally, the manual sync between wall clock time and timer time scares me.

PR #303, opened 2025-04-11
Thorlabs VCP port and 4D detector mask

RemiSoummer:
This is fixing an issue with the detector mask. This works but we will need to revisit all the 4D work on hardware once the interferometer is repaired, we had to interrupt testing because the interferometer is not usable (fringe visibility fluctuation) and need to be send back to AZ. 

Also sneaking in the Thorlabs fix and comments about the randomly changing VCP ports described in #304 (sorry for the sneak)


steigersg (dismissed):
Approved with the understanding that this is a bandaid fix to continue progress and issue #304 will need to be addressed for a more permanent solution

steigersg (approved):
re approving after a flake8 fix

PR #112, opened 2023-06-29
Fix datastream lifetime.

ehpor:
~This PR is for a potential bug in data stream lifetime.~ (see comment below)

Fixes https://github.com/spacetelescope/catkit2/issues/113.

PR #308, opened 2025-04-29
Be strict about shared memory lifetime

ehpor:
> [!CAUTION]
> This PR requires #272 to be merged before.

This replaces #112.

This PR is to explicitly prohibit creating a new shared memory object with a certain ID, when one already exists. This is not allowed on Linux/MacOS but was allowed on Windows until now (due to me not correctly checking the error output). This PR performs stricter error checks. Simultaneously, all error values are now checked and the exceptions now contain text containing the exact error that was encountered rather than a generic message.

### DIFF
--- a/catkit_core/SharedMemory.cpp
+++ b/catkit_core/SharedMemory.cpp
@@ -47,12 +47,18 @@ SharedMemory::~SharedMemory()
 	if (m_Buffer)
 	{
 #ifdef _WIN32
+		// Unmap the view of the file.
 		UnmapViewOfFile(m_Buffer);
+
+		// Close the file mapping object. For Windows, once all handles are closed,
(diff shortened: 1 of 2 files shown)


RemiSoummer:
Installed on HiCAT PC and the test is passing

ehpor:
> Installed on HiCAT PC and the test is passing

Did you mean to approve or comment?

RemiSoummer:
I meant to comment, working on it now

RemiSoummer:
also tested ok on MacOS.

RemiSoummer (approved):
looks good and working ok on Mac OS and windows.  
I'm working on adding a few tests but will do in another PR to avoid delaying

PR #310, opened 2025-05-01
Add a benchmark for latency for events.

ehpor:
> [!CAUTION]
> This PR requires #272 to be merged before.

This PR adds a benchmark for measuring latency for events. This is done for all pairs of cores on the machine. Some measured graphs below on my AMD Ryzen 9 5900X to indicate why core-to-core latency measurements are important to know about.

This PR also removes explicit `Lock()` and `Unlock()` functions on the `EventImplementation`s, since this interfered with wait methods when locks are used. This PR also fixes a bug in the spin lock event implementation.

Usage: `event_latency semaphore > latency_for_semaphore.csv`

## AMD Ryzen 9 5900X (12 cores, 24 threads, Windows)

image
image

## AMD Ryzen Threadripper PRO 5995WX (64 cores, 24 threads, Linux)

image
image

### DIFF
--- a/catkit_core/EventSpinLock.inl
+++ b/catkit_core/EventSpinLock.inl
@@ -41,7 +41,7 @@ inline void EventSpinLock::Wait(double timeout_in_sec, std::function<bool()> con
 					error_check();
 
 				if (timer.GetTime() > timeout_in_sec)
-					throw std::runtime_error("Waiting time has expired2.");
+					throw std::runtime_error("Waiting time has expired.");
 
 				i = 0;
 
--- a/catkit_core/EventSpinLock.inl
+++ b/catkit_core/EventSpinLock.inl
@@ -2,6 +2,7 @@
 
 #include <atomic>
 #include <cstddef>
+#include <thread>
 
 using EventSpinLock = EventImpl<EventImplementationType::SpinLock>;
 
@@ -25,7 +26,7 @@ inline void EventSpinLock::Wait(double timeout_in_sec, std::function<bool()> con
 	{
 		while (true)
 		{
-			if (m_SharedState->m_Counter.load(std::memory_order_relaxed) == current_counter)
+			if (m_SharedState->m_Counter.load(std::memory_order_acquire) != current_counter)
 				break;
 
 			if (++i == NUM_ITERATIONS_BETWEEN_CHECKS)
@@ -38,6 +39,9 @@ inline void EventSpinLock::Wait(double timeout_in_sec, std::function<bool()> con
 					throw std::runtime_error("Waiting time has expired.");
 
 				i = 0;
+
+				// Yield the thread to allow other threads to run.
+				std::this_thread::yield();
 			}
 		}
 	}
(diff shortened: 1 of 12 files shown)


catkit_core/EventSpinLock.inl
```diff
@@ -35,9 +41,12 @@ inline void EventSpinLock::Wait(double timeout_in_sec, std::function<bool()> con
 					error_check();
 
 				if (timer.GetTime() > timeout_in_sec)
-					throw std::runtime_error("Waiting time has expired.");
+					throw std::runtime_error("Waiting time has expired2.");
```
steigersg: think this is a typo

ehpor: You're right. More-so a debugging leftover, since I might've been trying to figure out which of the few exceptions with the same text was being triggered. Fixed in b87c305.

steigersg:
@ehpor I'm trying to run `event_latency semaphore > latency_for_semaphore.csv` from the command line and am getting

`-bash: event_latency: command not found`

How are executables in C run? I recompiled catkit2 on this branch so that shouldn't be the issue

ehpor:
All binaries are compiled when installing catkit2, in an isolated environment in a temporary directory to ensure reproducable builds. Right now, these binaries are not installed themselves, that's for a separate PR (it's not trivial, apparently). To run them for now, modify `pyproject.toml` with a new line in the `[build-system]` section: `build-dir = "build"`. Then, rebuild catkit2. Then, you should see the binaries in the `build/benchmarks` directory. Some operating systems have a slightly different folder structure, so you might have to look around in the build folder to find them.

Also, not all event types have an implementation for all operating systems:
* Semaphore: Windows only
* Condition variable: Linux & MacOS only
* Futex: Linux only
* Spinlock: Windows, Linux & MacOS

steigersg:
> All binaries are compiled when installing catkit2, in an isolated environment in a temporary directory to ensure reproducable builds. Right now, these binaries are not installed themselves, that's for a separate PR (it's not trivial, apparently). To run them for now, modify `pyproject.toml` with a new line in the `[build-system]` section: `build-dir = "build"`. Then, rebuild catkit2. Then, you should see the binaries in the `build/benchmarks` directory. Some operating systems have a slightly different folder structure, so you might have to look around in the build folder to find them.
> 
> Also, not all event types have an implementation for all operating systems:
> 
> * Semaphore: Windows only
> * Condition variable: Linux & MacOS only
> * Futex: Linux only
> * Spinlock: Windows, Linux & MacOS

hmmm adding that line to the `pyproject.toml` still doesn't result in a `build` directory being created, it should appear in the same location the installation is run from correct?

ehpor:
Whoops, it should be under `[tool.scikit-build]`. @steigersg

steigersg (approved):
Got the benchmark working on my Mac. 
Screenshot 2025-05-30 at 2 10 54 PM

Only comment is that the 
`"Thread affinity is not supported on this platform."`
message gets spit out for each point on MacOS which shouldn't block this getting merged considering it is a benchmark test, but would be improved if it just stated it once.

Besides that LGTM!

ehpor:
Oh, right. Yep, there is no way to set which core a thread is running on on MacOS. (As in, MacOS doesn't have an API for doing this.) So therefore, this benchmark, which tests core-to-core latency, isn't gonna give useful data. I forgot about that detail.

So, your graph, ran on MacOS, will be uniform with noise.

PR #311, opened 2025-05-02
Add an event based on futex

ehpor:
> [!CAUTION]
> This PR requires #310 to be merged before.

Linux only. There might be a way to do MacOS too, but I remember there being issues with it not having a timeout. There is no way to do this type of synchronization accross processes in Windows, since it only supports intra-process futex.

This PR also adds a test for all (implemented) events, rather than just the default implementation.

Results on AMD Threadripper 5995WX (64 cores). This is the fastest (non-spinlock) synchronization primitive that is implemented on Linux by a full microsecond.
image

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -47,5 +47,5 @@ jobs:
     - name: Run tests
       run: |
         conda activate ci-env
-        pytest
+        pytest -v
       shell: bash -l {0}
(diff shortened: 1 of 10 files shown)


ehpor:
Tests are deadlocking due to #310 not being merged yet. I'll wait for that one before merging this.

ehpor:
Ready for review but see the caution in the first post.

ehpor:
Nope, the deadlock is still there. I'll have to check this again on Linux.

ehpor:
Regrettably, my inner goldfish prevailed and I forgot to actually implement timeouts for the futex event. That's why it was deadlocking. Now fixed.

ehpor:
Rebased after #310 was merged.

steigersg (approved):
Ran this benchmark on the hicattrois linux machine without issue so seems good

Screenshot 2025-05-30 at 4 23 36 PM

ehpor:
@steigersg For reference: what processor? And what Linux version?

steigersg:
> @steigersg For reference: what processor? And what Linux version?

Intel Xeon Gold 6242R x 80 Processor and Ubuntu 24.04.2 LTS

PR #317, opened 2025-05-13
Fix power setting in simulated NKT and MCLS1 Services 

steigersg:
For the NKT,  the `power` (and `power_setpoint`) is given as a percentage of the total power provided by the input spectrum which is defined under `spectrum_fname` in the `simulator.yml`. The power should therefore be multiplied by 1e-2 in the `set_emission` and `set_power_setpoint` functions to reflect the percentage. 

For the MCLS1, however, the `power` is just the total power in the wavefront and should not be multiplied by the 1e-2 factor. 

Something that is still unclear me though is why in the simulated services the `power_setpoint` from the `services.yml` are used to set the power in the `set_emission` and `set_power_setpoint` functions. Typically on hardware we work with currents and so I'm not even sure if the `power` from the `simulator.yml` and the `power_setpoint` from the `services.yml` are referring to the same thing here.

### DIFF
--- a/catkit2/services/nkt_superk_sim/nkt_superk_sim.py
+++ b/catkit2/services/nkt_superk_sim/nkt_superk_sim.py
@@ -114,7 +114,7 @@ class NktSuperkSim(Service):
     def set_power_setpoint(self, power_setpoint):
         self.testbed.simulator.set_source_power(
             source_name=self.id,
-            power=self.emission.get()[0] * power_setpoint
+            power=self.emission.get()[0] * power_setpoint * 1e-2
         )
 
     def set_current_setpoint(self, current_setpoint):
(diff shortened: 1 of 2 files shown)


ehpor:
The `testbed.simulator.set_source_power()` sets the total power of the wavefront in the simulator. It is the job of the simulated service to translate between the hardware commands and the resulting total power of the wavefront. The NKT can be switched between current and power mode on hardware. Technically, the simulated NKT service should keep track of what mode it is in and not send the power to the simulator if it's not in that mode.

FYI: I never did the current->power calibration, since we rarely switched power in the simulator, so that's why the current setpoint was not implemented. And since we never modify NKT power/current in simulation, we didn't care.

RemiSoummer (approved):
We are going to merge the bug fix here, and come back to this in more detail to remove the confusion between current or power, and percentage or not, and how the power should be calibrated in the simulation. 

Following this up in #318

 This will be necessary for HiCAT side when we add our new source.    This will definitely be clarified on the hicat-package2 side here: https://github.com/spacetelescope/hicat-package2/issues/712

PR #320, opened 2025-05-15
Move lock upstream to the deformable mirror base class.

### LINKED ISSUE #319: DM dropping commands if apply shape back to back
When sending commands to two different channels back to back:

```
dm.apply_shape('aberration', (sinx_cmd.shaped, np.zeros((34, 34))))
dm.apply_shape('poke', (siny_cmd.shaped, np.zeros((34, 34))))
```

In ~3-6% of the cases, the first command (sinx_cmd here) is not applied. The second one is always applied. 

This problem can be solved by applying a delay of ~4-5 ms between the commands. 2 ms is not sufficient (improves but not sufficient). 
Note: the delay is applied by dropping frames on the Zernike camera at 500 Hz, which is more reliable than time.sleep (unusable under 10 ms). 

A code example that reproduces the problem on HiCAT can be found here: https://github.com/spacetelescope/hicat-package2/blob/feature/analytical_zwfs_measurements/sandbox/rpourcelot/dm_glitch.ipynb

ehpor:
Fixes #319.

### DIFF
--- a/catkit2/base_services/deformable_mirror.py
+++ b/catkit2/base_services/deformable_mirror.py
@@ -37,6 +37,8 @@ class DeformableMirrorService(Service):
         self.total_voltage = self.make_data_stream('total_voltage', 'float64', [self.dm_command_length], 20)
         self.total_surface = self.make_data_stream('total_surface', 'float64', [self.dm_command_length], 20)
 
+        self.lock = threading.Lock()
+
     @property
     def dm_command_length(self):
         '''The command length of the DM(s).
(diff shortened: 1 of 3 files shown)


ehpor:
@raphaelpclt @RemiSoummer I'm unable to test this, since we're using an independent C++ based control for our BMC DMs.

(FYI: that C++ based service uses a different way of adding together the channels and is therefore inherently invulnerable to this race condition.)

RemiSoummer:
Tested on hardware on HiCAT  by running https://github.com/spacetelescope/hicat-package2/blob/feature/analytical_zwfs_measurements/sandbox/rpourcelot/dm_glitch.ipynb
This essentially applies two commands (a sin_x and sin_y back to back) and checked whether the results look like sin_x, sin_y or the sum of both.   

```
        # Apply two commands sequentially
        dm.apply_shape('aberration', (sinx_cmd.shaped, np.zeros((34, 34))))
        dm.apply_shape('poke', (siny_cmd.shaped, np.zeros((34, 34))))
``` 

On dev we have about 4% failure (tested on 1000 realizations) and zero failure rate on the PR. 

The issue is the the **first** command is not applied when this issues manifests, and only the **second** command is applied,  which is very weird and hard to explain.

RemiSoummer (approved):
Tested on hardware and this fixes the issue.  Now it is not entirely clear what is actually happening and clearly some very nasty race condition.

RemiSoummer:
@ehpor leaving open for now in case you want to add more thoughts as to why this is working... but go ahead and merge if you want.

ehpor:
If it fixes the problem, it's definitely a race condition. I don't think not knowing the exact race condition is enough reason to block this from merging. Merging now.

raphaelpclt:
Thanks @ehpor  for the quick solve! One thought - but not urgent at all - does this fix sacrifice speed? Looking at the trace now, there is no overlap of the commands. Could it be slightly more optimized?  
image

ehpor:
@raphaelpclt A tiny bit. Before, the `compute_xxx` (I don't remember what I named it) and part of the `send_surface` (the part before the `send_data`) could overlap, but the `send_data` itself could not (since it solely consists of a BMC API call that we want to  perform sequential since the API is not thread-safe, at least I'm not 100% sure it's thread-safe).

So, yes, there is a small impact on performance, but since the `send_data` is the majority of the effort of the DM, it's not much.

PR #322, opened 2025-05-16
Cleanup callbacks and properties after a `Server` or `Service` stops.

### LINKED ISSUE #321: `Testbed` and `Service` are not destructed at the end of the program
The reference count after `Testbed.run()` is 9, which is 1 for the variable itself and 8 for the functions that are used all callbacks by `Server`. A `Service` has the same problem, not only with the `Server` but also with `Service.make_command()` and `Service.make_property()` that each use callbacks.

These callbacks should use weak references instead, which are make strong references right before they are used.

ehpor:
Fixes #321.

### DIFF
--- a/catkit2/__init__.py
+++ b/catkit2/__init__.py
@@ -1,8 +1,12 @@
 # flake8: noqa
 
+# Setting to ensure CTRL-C commands are caught, which allows services to exit properly.
 import os
+os.environ['FOR_DISABLE_CONSOLE_CTRL_HANDLER'] = '1'
 
+# Enable printing of stacktrace upon segfault.
 import faulthandler
+faulthandler.enable()
 
 from . import testbed
 from . import simulator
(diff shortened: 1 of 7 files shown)


ehpor:
The proposed solution doesn't work for lambdas that are passed as callbacks. Since these are temporary functions, without storing a strong reference, they are garbage collected right after being passed as a callback.

ehpor:
Reworked the solution to instead cleanup all callbacks and attributes after stopping the service/server. Also added protection against running the same server/service twice. (This would start the service without API or the server without request handlers, which would feel very weird to the user. Better to have an explicit error.) Neither should be used anyways, but is currently allowed.

I also moved the fortran environment variable to the top. This shouldn't make any difference, but Fortran on my PC clobbered the Ctrl+C handler without it there. 🤷

ehpor:
Added a test for `Server` cleanup. Note that the `Service` cleanup is not tested. Doing this in an automated fashion is not trivial, so I'm not attempting to do it here.

Checked that the test fails for the current develop and passes on this PR.

RemiSoummer (approved):
Looks good, and ran the test successfully.

steigersg:
Also ran the test successfully (unsuccessful on dev)

PR #323, opened 2025-05-17
Add a message broker inside a testbed

ehpor:
> [!CAUTION]
> This PR requires #322 to be merged before.

This PR adds a `MessageBroker` instance to a `Testbed` instance and allows access to it via a `TestbedProxy`.


RemiSoummer:
@ehpor could you add a pytest or a basic demo notebook so we can start testing things?

RemiSoummer:
actually much less in here than it seemed now that we have merged #322

ehpor:
Yep, this doesn't seem ready. The bindings are missing. Working on it now.

ehpor:
> @ehpor could you add a pytest or a basic demo notebook so we can start testing things?

This is only adding a `MessageBroker` object to any testbed that is ran. It just means that after this PR, I can start testing myself. A message broker isn't really meant to access directly, yet. There is nothing that gets published to it right now. Also, the exact API is quite flexible right now, since I'm trying to find the easiest way to use it.

ehpor:
> Yep, this doesn't seem ready. The bindings are missing. Working on it now.

Done. @RemiSoummer

RemiSoummer:
ok thanks, we are testing that with @steigersg - both recompiled and we can create a TestbedProxy and there is indeed a message broker and we can access it.  For example, we can get an empty array back from `testbed.message_broker.get_all_message_topics()`

Let us know if there is anything else you'd like to test (and if you can add some kind of test or notebook demo for that)

ehpor:
That's it. That's all that this PR does: it just adds a `testbed.message_broker` attribute. All `MessageBroker` functionality that the tests already test for, should work on this exposed message broker. @RemiSoummer 

Future PRs add functionality that automatically publish a lot of information on it. Like CPU/memory usage of services. Then tracing. And later all properties/datastream data.

RemiSoummer (approved):
ok let's merge this since we checked that this is working well.

Issue #304, opened 2025-04-11
Migrate hardcoded VCP port in MCLS1 service


steigersg:
There is a hard coded VCP port in the MCLS service that is system specific since it describes which port the MCLS is connected to. 

See comments in PR #303 

This should be moved to the services.yaml (unless I am mistaken and we should always expect the MCLS1 to be on the same VCP port for all systems that are using catkit2).

ehpor:
Any reason why we're using UART directly rather than using the Thorlabs Python SDK for the MCLS1?

raphaelpclt:
> Any reason why we're using UART directly rather than using the Thorlabs Python SDK for the MCLS1?

I think I just copied catkit1 one for that if I recall correctly

ehpor:
> I think I just copied catkit1 one for that if I recall correctly

That's totally fine. I just wanted to make sure there wasn't something wrong with the SDK in case we want to switch over in the future for some reason.

RemiSoummer:
We'll come back to this, especially when we implement the new Thorlabs laser, so for now keep chasing the port number with #326

RemiSoummer:
We found another side effect of this problem with @steigersg that we had not seen before.  In the past the VCP port had always jumped to an unused port, so testing for the existence of a connection on the expected port was sufficient.  

This time around the VCP port switched to a port that was actually used by something esle. A a result the behavior was that it was a not possible to interact with the source, and we were not getting the COM/VCP error message. 
Instead the issue appeared as a submission of an empty string to the mcls1 data stream which "string failed to convert to float" (line 79 of thorlabs_mcls1.py). 

PR #397 adds a test to catch the ValueError and also updates the port to the current value.

PR #326, opened 2025-05-20
updating the VCP port

RemiSoummer:
The VCP port of the MCLS1 keeps changing by itself after rebooting, so this PR updates it to be consistent with the current hardware state. 

This needs to be fixed properly later because this is a HiCAT hardware dependency. 

We still need to get back to #304 to solve this properly


steigersg (approved):
Looks good to me - will need to revisit the best way to handle port setting in the future though to make this not testbed specific (or move the port value to a config)

Issue #327, opened 2025-05-20
Boston DM Base Class Refactor Bug with PCIe device ID


steigersg:
See also discussion in #287 

`self.dm` should be `self.device` in the `RuntimeError`s in the `open` function of `bmc_deformable_mirror_hardware.py`.  This is addressed in #328 

Additionally the `set_device_id` function does not handle being passed a 0 (the default) gracefully and instead raises an error. This is a problem for HiCAT which does not specify a `device_id`.

PR #328, opened 2025-05-20
Hack for BMMC DMs on HiCAT with PCIe device setting 

steigersg:
Partially addresses #327. 

This is a hack to allow the DMs to work on HiCAT after the DM refactor. Another future PR will need to address the underlying issue of why the `set_device_id` function is not handling being passed 0 in the expected way.  

- [x] Tested on HiCAT hardware  

@ivalaginja would appreciate your input/testing to make sure this doesn't impact anything with THD2


ivalaginja (approved):
We had talked about this a while ago, I agree this is the way to proceed.

PR #329, opened 2025-05-25
C++ deformable mirror base class

ehpor:
Work in progress. Todo:

- [x] Add startup maps.
- [x] Adding and updating data streams.
- [x] Add properties.
- [ ] other things?

This PR adds a C++ base class for deformable mirrors. This allows us to perform multiple channel updates asynchronously and independently, unlike the current Python base class. This base class handles channels, initialization of channels and updating from channels, feeding into a generic `UpdateSurface()` function call, to be implemented by child classes. Rather than adding all channels directly, this class performs delta updates on the channels, allowing calculation of the delta to be performed asynchronously between channels. Afterwards, this delta has to be applied synchronously to the DM.

> [!NOTE]  
> Due to floating point arithmetic, the delta updates will not be the same as absolute updates (which would add all channels explicitly), and errors will accumulate over time. However, the errors are negligible. Let's do a computation to figure out how much. With 64-bit numbers we have a 52-bit mantissa. This gives a $2^{-53}\approx10^{-16}$ relative error per floating point operation. So, after a month of continuous operation at 10kHz, we would expect a $\sqrt{10000\times60 \times60\times24\times30}\times10^{-16}\approx10^{-11}$ relative error, equal to 36-bit, far less than our DMs.

This PR also adds functionality for reading FITS files to the C++ core.

### DIFF
--- a/catkit_core/DeformableMirrorService.cpp
+++ b/catkit_core/DeformableMirrorService.cpp
@@ -1,0 +1,134 @@
+#include "DeformableMirrorService.h"
+
+#include "FitsFile.h"
+
+template<typename T>
+void CastToBool(T *data, bool *mask, size_t size)
+{
+	for (size_t i = 0; i < size; ++i)
+		mask[i] = data[i] > 0;
+}
+
+template<typename T>
+void CastToDouble(T *data, double *mask, size_t size)
+{
+	for (size_t i = 0; i < size; ++i)
+		mask[i] = data[i];
+}
+
+DeformableMirrorService::DeformableMirrorService(std::string service_type, std::string service_id, int service_port, int testbed_port)
+	: Service(service_type, service_id, service_port, testbed_port)
+{
+	// Get the actuator mask.
+	auto config = GetConfig();
+
+	if (!config.contains("device_actuator_mask_fname"))
+		throw std::runtime_error("Need to provide a device actuator mask for the deformable mirror.");
+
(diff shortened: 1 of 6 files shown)


ehpor:
This is for incorporating a C++ BMC service, which would simplify #85 (ie. upgrading to be independent of Python version).

ivalaginja:
This is exclusively adding code, so the stakes are very low. Have you been able to test this in some way yet or would more in-depth testing happen with a future implementation of a child class?

ehpor:
It's a generalization refactor of the BMC implementation on SEAL. But I haven't done any testing on this specific version. I'm expecting to find 1-2 bugs/typos when moving over the BMC implementation itself. Same for the ALPAO.

I'm just trying to keep the number of lines limited per PR. ~500 lines is already a lot of C++ code to review, even when it is just adding isolated code.

ivalaginja (commented):
With my limited C++ knowledge, I did my best to review this and see no issues.

When running Copilot on this PR, one returned point stood out, which is: "[...] the `status` variable is declared but not initialized to zero. According to cfitsio documentation, status must always be set to 0 before passing it to these functions."

This seemed like something to address, but you might be able to tell more immediately than me whether the above is true or not @ehpor.

ehpor:
@ivalaginja That is 100% correct. I just checked the docs. I'll fix that.

ehpor:
@ivalaginja Fixed in 941662e.

ivalaginja (approved):
LGTM

PR #331, opened 2025-05-29
Add ArrayView type for message broker.

ehpor:
The array view will ultimately replace `Tensor`. The `ArrayView` class mimics a Numpy array, but does not own the data. It is purely meant as a way to transfer data between Python and C++, and to provide good type information for arrays in C++ without binding ourselves to a specific math framework.

An `ArrayView` does not own it's data, while a `Tensor` could (depending on the given parameters). The uncertainness of ownership led to a few memory leaks before (fixed long ago), but there still could be a few out there. As an `ArrayView` doesn't own its data, it is difficult to make those mistakes again.

Compared to `Tensor`, an `ArrayView` can contain values that are big/little endian, and booleans (in addition to the other dtypes).

While `Tensor` was included as part of the `Value` variant object, this is not intended to be the case for `ArrayView`. Inclusion means that it can be a `Property` of a `Service`, a feature that was never used. That requirement meant that in some cases a `Tensor` could own its data. As a result, an `ArrayView` cannot be a `Property` of a `Service` anymore (until the MessageBroker refactor, that is), as the name implies.


catkit_core/MessageBroker.h
```diff
@@ -25,7 +25,6 @@ const size_t HOST_NAME_SIZE = 64;
 const size_t METADATA_MAX_STRLEN = 8;
 const size_t METADATA_MAX_KEYLEN = 7;
 const size_t MAX_NUM_MESSAGES = 65536;
-const size_t MAX_NUM_DIMENSIONS = 4;
```
steigersg, line 28: am I correct that this was removed because its now in the header of the ArrayView?

ehpor: Yes.

tests/test_message_broker.py
```diff
@@ -76,6 +76,21 @@ def test_message_dtype_and_shape(broker, shape, dtype):
     assert retrieved_message.payload.ndim == len(shape)
     assert retrieved_message.topic == topic
 
+def test_message_broker_publish_array(broker):
+    arr = np.random.randn(10)
+    broker.publish_array('test_array', arr)
+
+    msg = broker.get_newest_message('test_array')
+    assert np.allclose(msg.payload, arr)
```
steigersg: Why is `np.allclose` used here? Is it just to avoid float rounding/precision issues in the test or do we expect the message broker to potentially impact the values of the array?

ehpor: Fixed in 350682f.

steigersg (approved):
After the explanations/small fixes this now looks good to me!

PR #332, opened 2025-05-30
updates to shared memory tests

RemiSoummer:
Following #308 I was doing some tests so thought I might as well have a quick PR to add add a bit more testing here. 
Not sure how useful that will be here.

### DIFF
--- a/tests/test_shared_memory.py
+++ b/tests/test_shared_memory.py
@@ -6,9 +6,10 @@ def test_shared_memory_lifetime():
     stream_id = 'test_memory'
     memory = SharedMemory.create(stream_id, 1024)
 
-    # Creating shared memory with the same ID should raise en erorr.
-    with pytest.raises(RuntimeError):
+    # Creating shared memory with the same ID should raise an error.
+    with pytest.raises(RuntimeError) as excinfo:
         SharedMemory.create(stream_id, 1024)
+    assert "File exists" in str(excinfo.value)
 
     # We should be able to access the shared memory when opening.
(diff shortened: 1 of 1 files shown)


ehpor:
I'd use `pytest.raises(RuntimeError, match='string to match to the exception string')` instead of an assert after the context manager. See https://docs.pytest.org/en/stable/reference/reference.html#pytest-raises

RemiSoummer:
@ehpor also I was trying to be fancy can catch the specific error message, but that is not working on Windows.  So I reverted that to the base message from `SharedMemory.cpp`. 
Also, for the `TypeError` I don't have the full string so `match`  does not work, and for now I reverted to the previous syntax. 
If you have another suggestion I can implement, otherwise we can skip that test if you prefer. 

Unfortunately there is no `contains=` option in `pytest.raises`

ehpor:
@RemiSoummer `match` is a full-on regex string. So it matches substrings in the exception string. An example from the docs:

```python
with pytest.raises(ValueError, match='must be 0 or None'):
    raise ValueError("value must be 0 or None")
```

RemiSoummer:
Ok thanks - for some reason I must have done the wrong thing and though match was not working.

RemiSoummer:
ha but not that doesn't work on windows, so can't do here, back to filtering on generic message

ehpor (approved):
LGTM. And you fixed my typo in line 9 as well XD

FYI: your hardship is the reason why people usually don't bother testing for specific error messages. It makes the tests fragile.

PR #334, opened 2025-05-30
Add a `Service` class that communicates solely via the `MessageBroker`.

ehpor:
> [!CAUTION]
> This PR requires #331 to be merged before.

The `ServiceV2` class is supposed to be a drop-in replacement for the current `Service` class in most cases. However, this will require extensive testing before making this switch. This PR just adds the class and a compile-time switch to allow testing of this feature.


a-sevin:
I'm trying to test it but `Property.h` is missing in the repository:

```log
catkit_core/ServiceV2.h:12:10: fatal error: Property.h: No such file or directory
         12 | #include "Property.h"
            |          ^~~~~~~~~~~~
```

ehpor:
Reworked into several other PRs. This PR is no longer needed.

PR #335, opened 2025-06-06
Recreate shared memory for message broker if it already exists.

ehpor:
In some cases, the shared memory used by the message broker already exists.  In this case, the creation of a shared memory block will fail. This is intended: 

This is in cases where the shared memory block was somehow not destroyed correctly after the process shut down. For example when the testbed crashes or somehow cleanup wasn't performed. In these cases, we're left with a dangling shared memory block that prevents the testbed from starting again.

This PR uses a file lock (a lock which gets unlocked automatically, even when the process exits prematurely) to establish whether it can reuse the shared memory block. If this lock is not held, a new testbed server will acquire it and afterwards be allowed to recreate a shared memory block, even if it already exists.

### DIFF
--- a/catkit2/bindings.cpp
+++ b/catkit2/bindings.cpp
@@ -891,6 +891,7 @@ PYBIND11_MODULE(catkit_bindings, m)
 			auto mem = SharedMemory::Open(name);
 			return std::shared_ptr<SharedMemory>(std::move(mem));
 		})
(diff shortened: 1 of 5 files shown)


ivalaginja (approved):
LGTM. Tested on my Mac 12.2 and I can start a (THD2) server all right.

PR #338, opened 2025-06-11
Add Phasics SID4 Service to Catkit2

### LINKED ISSUE #337: Add Phasics SID4 Service to catkit2
I just received permission from Phasics to provide our Phasics SID4 service to catkit2.

This service requires a custom Python package named PySID4x which must be acquired separately from the manufacturer.

Thoughts @RemiSoummer @ehpor @steigersg @raphaelpclt  ?

lanemeier7:
Adding Phasics SID4 camera service to catkti2. Name of service is phasics_cam

Phasics Camera
=========================

This service operates a Phasics SID4 camera using the Phasics SDK for communication and PySID4x python library. It handles image acquisition, and by default returns Phase Maps as images. For raw intensity values see the 'intensity' data stream.

>[!NOTE]
>The Phasics Camera requires the Phasics SDK and PySID4x Python package to be installed and configured correctly. To get a copy of the Phasics SDK and PySID4x python package please contact Phasics directly for customer support.

Configuration
-------------

``` yaml
phasics_cam:
  service_type: phasics_cam
  simulated_service_type: phasics_cam_sim
  interface: camera
  mode: normal
  mask: !path "../../data/phasics/2025-01-28 full aperture.msk"
  usr_profile_path: C:/path/to/SID4_SDK_x86_64/Examples/UserProfiles/SID4-1481/SID4-1481.txt
  wavelength: 633.0
  buffer_frames: 16
  requires_safety: false
  exposure_time: 504
```

``exposure_time``: Exposure time in milliseconds.

``mode``: 'normal' mode operates the camera. 'dummy' mode generates fake data for testing signal chain.

``wavelength``: The wavelength in (nm) of the camera

``mask``: the location of the camera mask

``usr_profile_path``: the location of hte SID4 SDK user profile

``buffer frames``: number of buffered frames in the catkit2 datastream

Commands
-----------
``take_measurement``: Acquires a single measurement from the Phasics camera.

``start_acquisition``: Begins continuous image acquisition from the camera.

``end_acquisition``: Stops the continuous image acquisition.

``phase_filtering``: Applies phase filtering to the acquired images using supplied zernike projection coefficients

``phase_projection``: Performs zernike phase projection on the acquired images.

Datastreams
-----------
``images``: The acquired (Phase map) images from the Phasics camera.

``is_acquiring``: Indicates whether the service is currently acquiring data (True for acquiring, False for not acquiring).

``intensity``: The acquired intensity images from Phasics camera.


catkit2/testbed/testbed.py
lanemeier7, line 1: @ehpor I didn't mean to contribute this file on this go around. But I need this change to get Python 3.8 working with catkit2.

ehpor: Interesting. I'm running 3.8 on SEAL and it's working fine. Also, I'm running 3.11 on my dev machine and it's working fine there too. I wonder what's wrong.

lanemeier7: Our machines are using Windows 10 & 11 and Python 3.8.20 produces this error:

``` python
server = Testbed(port=port, is_simulated=is_simulated, config=config)
  File "C:\catkit2\catkit2\testbed\testbed.py", line 282, in __init__
    module = entry_point.module
AttributeError: 'EntryPoint' object has no attribute 'module'
```

lanemeier7: I think specifically in Python version 3.8  they don't have a 'module' only a 'name'. It does look like the module term returns in later versions...

``` python
EntryPoint(name='phasics_cam', value='catkit2.services.phasics_cam.phasics_cam', group='catkit2.services'),
```

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+class PhasicsCam(Service):
+    """
+    A service class for the Phasics camera.
+
+    A camera service with functionality for taking measurements,
+    starting and stopping live acquisitions, and managing an image buffer.
+
+    Parameters
+    ----------
+    Service : catkit2.testbed.service.Service
```
ehpor: This is not part of the parameters. The `__init__()` has no parameters so this section shouldn't be here.

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+        self.log.info(f'loading mask {maskname}')
+        SID4Session.change_mask(filename=maskname)
+        SID4Session.Camera.load()
+        SID4Session.Camera.set_attribute('Exposure', self.exposure_time)
+        SID4Session.modify_user_profile(wavelength=self.config.get('wavelength'))
+        SID4Session.Zernike.open()
+        poly_order = 9
+        SID4Session.Zernike.update_projection(poly_order, source='user profile')
+
+        self.session = SID4Session
```
ehpor: Any reason for not putting `SID4Session` in `self.session` right away?

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+        num_frames_in_buffer : int
+            The number of frames to maintain in the data buffer.
+        images : catkit2.testbed.service.DataStream
+            A data stream to manage phase map image frames.
+        intensity : catkit2.testbed.service.DataStream
+            A data stream to manage intensity image frames.
+        """
+        super().__init__('phasics_cam')
+        # in dummy mode fake data is generated by API or device
+        SID4Session = self.initialize_sid4_session(mode=self.config.get('mode', 'normal'))
```
ehpor: Variable names should follow Python naming conventions. Ie. `sid4_session`.

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+        self.is_acquiring = self.make_data_stream('is_acquiring', 'int8', [1], self.num_frames_in_buffer)
+
+        # create commands
+        self.make_command('take_measurement', self.take_measurement)
+        self.make_command('start_acquisition', self.start_acquisition)
+        self.make_command('end_acquisition', self.end_acquisition)
+        self.make_command('phase_filtering', self.phase_filtering)
+        self.make_command('phase_projection', self.phase_projection)
+
+        self.framecounter = 0
```
ehpor: This variable is not used?

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+
+                has_correct_parameters = np.allclose(self.intensity.shape, intensity.shape)
+
+                if not has_correct_parameters:
+                    self.intensity.update_parameters('float32', intensity.shape, self.num_frames_in_buffer)
+                self.intensity.submit_data(intensity.astype('float32'))
+        finally:
+            session_name.Camera.stop_live_mode()
+            self.is_acquiring.submit_data(np.array([0], dtype='int8'))
+
```
ehpor, line 280: Only a single blank line here. Same for after this function.

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
+# -*- coding: utf-8 -*-
```
ehpor: Let's remove this preamble. Authorship is tracked via GIT. The formatting of the docstring can vary so it's not machine readable anyways. Same for the other file.

catkit2/testbed/testbed.py
```diff
@@ -309,11 +309,20 @@ def __init__(self, port, is_simulated, config):
...
         self.service_type_paths = {}
-        for entry_point in importlib_metadata.entry_points()["catkit2.services"]:
-            module = entry_point.module
+        for entry_point in importlib_metadata.entry_points().get("catkit2.services", []):
+            try:
+                # For Python 3.8+
+                module = entry_point.value.split(':')[0]
+            except AttributeError:
+                # Fallback for older versions
+                module = entry_point.module
```
ehpor: I'd turn this around. The workaround is now the main path. Exceptions should be exceptional.

docs/services/phasics_cam
```diff
@@ -0,0 +1,68 @@
+Phasics Camera
```
ehpor: This should be an `.rst` file like the others.

docs/services/phasics_cam
```diff
@@ -0,0 +1,68 @@
...
+
+.. code-block:: YAML
+
+phasics_cam:
+  service_type: phasics_cam
+  simulated_service_type: phasics_cam_sim
+  interface: camera
+  mode: normal
+  mask: !path "../../data/phasics/2025-01-28 full aperture.msk"
+  dll_path: C:/path/to/SID4_SDK_x86_64/lib
```
ehpor: I'd prefer to have this an environment variable instead. That's more how the other services do it. Say: `CATKIT_PHASICS_ROOT` that points to the SDK path.

ehpor (changes requested):

ehpor:
Also, use rebases for updating your branch to be up-to-date with develop rather than merge commits. This avoids many superfluous merge commits.

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+            The width of the images in pixels.
+        num_frames_in_buffer : int
+            The number of frames to maintain in the data buffer.
+        images : catkit2.testbed.service.DataStream
+            A data stream to manage phase map image frames.
+        intensity : catkit2.testbed.service.DataStream
+            A data stream to manage intensity image frames.
+        """
+        super().__init__('phasics_cam')
+        # in dummy mode fake data is generated by API or device
```
ehpor, line 65: Most of the following code shouldn't be in the `__init__()`, but in the `open()`. While in the nominal case there is no difference, if there is an exception thrown, error handling is not performed in the same way. Typically, in the `__init__()` nothing should ever fail if there is a problem with the device itself.

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+        ----------
+        num_images : int, optional
+            The number of exposures to average. The default is 1.
+
+        Returns
+        -------
+        numpy.ndarray
+            The phase map (single or averaged) of dimensions (image_height, image_width).
+        """
+        session_name = self.session
```
ehpor: I don't understand this renaming? They are the same number of characters to type.

catkit2/services/phasics_cam/phasics_cam.py
```diff
@@ -0,0 +1,286 @@
...
+        SID4Session.Zernike.open()
+        poly_order = 9
+        SID4Session.Zernike.update_projection(poly_order, source='user profile')
+
+        self.session = SID4Session
+
+        # take an image to determine the size of the frames based on the mask provided
+        self.log.info('getting test image')
+        self.set_exposure_time(self.exposure_time)
+        test_image = self.take_measurement()
```
ehpor: I don't like having to take a measurement to know the size of the output images. Note: you can create a datastream with the shape of the largest image that you'll take (ie. it doesn't need to be the exact size) and then reshape the datastream in the `open()` based on a temp measurement. I'd put the `max_width` and `max_height` in the config.

docs/services/phasics_cam
```diff
@@ -0,0 +1,68 @@
...
+  usr_profile_path: C:/path/to/SID4_SDK_x86_64/Examples/UserProfiles/SID4-1481/SID4-1481.txt
+  wavelength: 633.0
+  buffer_frames: 16
+  requires_safety: false
+  exposure_time: 504
+
+Properties
+-----------
+``exposure_time``: Exposure time in milliseconds.
```
ehpor: These properties are not exposed by the service. There are no `self.make_property()` calls anywhere.

ehpor (changes requested):
And now also the main code, now that I'm home.

docs/services/phasics_cam
lanemeier7: I was thinking this would be a good way to describe the configuration parameters. What is a good way to describe it instead?

ehpor: Put that in the configuration section. The properties section is for the properties that are exposed by the service.

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: I've moved this to the class attributes section. Parameters is properly labeled as None

https://github.com/spacetelescope/catkit2/blob/774dc9229b818b5846aed61c165e8c398f03c5a9/catkit2/services/phasics_cam/phasics_cam.py#L28-L52

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: No reason. This is now being put into the SID4 session right away and the self.sid4_session is used instead of the extra variable.

https://github.com/spacetelescope/catkit2/blob/774dc9229b818b5846aed61c165e8c398f03c5a9/catkit2/services/phasics_cam/phasics_cam.py#L66-L79

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: This has been renamed to `self.sid4_session` using standard Python naming convention

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: The variable was not being used and is now removed.

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: Blank line removed.. not sure how to point to lines from commits in conversations.

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: Done

catkit2/testbed/testbed.py
lanemeier7: How is this?

https://github.com/spacetelescope/catkit2/blob/774dc9229b818b5846aed61c165e8c398f03c5a9/catkit2/testbed/testbed.py#L310-L326

docs/services/phasics_cam
lanemeier7: Done

docs/services/phasics_cam
lanemeier7: I went with `SID4_SDK_x86_64_LIB` because the pre requisite SDK will add this variable to the PATH. So no extra steps once that is installed. Then I followed the same pattern I saw for zwo_camera.py

https://github.com/spacetelescope/catkit2/blob/774dc9229b818b5846aed61c165e8c398f03c5a9/catkit2/services/phasics_cam/phasics_cam.py#L10-L22

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: This has been moved to `open`

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: renamed to session for convenience

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: Done! now using max_width / max_height instead

docs/services/phasics_cam
lanemeier7: Descriptions have been moved to configuration section, see cade02c9048c2eb8ac673401632a200c90ab78e4

catkit2/testbed/testbed.py
ehpor: Perfect.

catkit2/services/phasics_cam/phasics_cam.py
lanemeier7: Is the following OK for convenience? Or would you prefer self.session?

https://github.com/spacetelescope/catkit2/blob/cade02c9048c2eb8ac673401632a200c90ab78e4/catkit2/services/phasics_cam/phasics_cam.py#L260-L266

catkit2/testbed/testbed.py
lanemeier7: I'm also still confused about this one. Looking at https://docs.python.org/3.8/library/importlib.metadata.html#entry-points these entry points look like mine..  

`EntryPoint(name='wheel', value='wheel.cli:main', group='console_scripts')` 

Perhaps this is a change in Python 3.8.2 but in 3.8.0 its the old method. I'm struggling to find the Python 3.7 documentation on this

catkit2/services/phasics_cam/phasics_cam.py
ehpor: I'm fine with this.

catkit2/testbed/testbed.py
ehpor: Yep, double checked. I'm running the existing code on Python 3.8.20 on SEAL with importlib_metadata 7.0.1.

I don't think it matters. You're not breaking anything with putting this in. I'm fine with this change.

ehpor (approved):

lanemeier7:
@ehpor @steigersg This is ready to merge now thanks or reviewing! Which merge strategy should I use for this PR?

ehpor:
@lanemeier7 Just the normal "Create a merge commit". The rebasing is just for keeping your branch up-to-date with the develop branch before/during/after review. For the actual merge, we chose to use normal merge commits.

PR #340, opened 2025-06-11
Fix for HybridPoolAllocator wrongly removing pools.

ehpor:
When the last slot on a pool was released, the pool will be released. This was done regardless of whether the pool was owned by the process or not, leading to a segmentation fault or bus error when the process didn't own the pool.

Pools are allocated blocks on the BuddyAllocator. Each pool has 64 slots, which the HybridPoolAllocator hands out. Each of the slots are reference counted by the HybridPoolAllocator. Slots are deallocated once the reference count reaches zero. Each of the pools are reference counted by the BuddyAllocator. The number of references to a Pool is 1 + N_allocated_slots, so at most 65. This PR adds proper reference counting for the Pools, which was done incorrectly.

Todo:
- [x] The pools of destroyed allocator objects should also be deallocated automatically after all references are released.

### DIFF
--- a/catkit_core/HybridPoolAllocator.cpp
+++ b/catkit_core/HybridPoolAllocator.cpp
@@ -3,6 +3,10 @@
 
 #include <array>
 #include <stdexcept>
+#include <iostream>
+
+//#define DEBUG_PRINT(a) std::cout << a << std::endl
+#define DEBUG_PRINT(a)
 
 const std::array<std::uint8_t, 4> HYBRID_POOL_ALLOCATOR_VERSION = {0, 0, 0, 0};
 
@@ -66,6 +70,8 @@ std::size_t HybridPoolAllocator::GetSharedStateSize(std::size_t capacity, std::s
 
 HybridPoolAllocator::Handle HybridPoolAllocator::Allocate(std::size_t size)
 {
+	DEBUG_PRINT("HybridPoolAllocator::Allocate(" << size << ")");
+
 	size = round_up_to_power_of_2(size);
 
 	if (size < m_MinSize)
@@ -73,6 +79,8 @@ HybridPoolAllocator::Handle HybridPoolAllocator::Allocate(std::size_t size)
 
 	if (size >= m_MinSizePool)
 	{
+		DEBUG_PRINT("size is too large for pools");
+
 		// The size is too large to use pools. Defer to the buddy allocator.
 		return m_Allocator->Allocate(size);
 	}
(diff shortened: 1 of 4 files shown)


ehpor:
~~Tests pass, but I feel there's still something wrong.~~ Nope, that was me making rebase mistakes.

It's ready for a first pass of review though.

catkit_core/HybridPoolAllocator.cpp
```diff
@@ -190,22 +260,46 @@ bool HybridPoolAllocator::Release(Handle handle)
...
+
 			bucket.emplace_back(pool_handle);
 		}
 
 		// Update the pool bitmap.
 		pool.map &= ~mask;
 
+		// Decrement the ref count of the pool to account for the slot being released.
+		m_Allocator->Release(pool_handle);
+
```
alexisyslau, line 273: _If I understand this:_

Before: Unconditionally releases the pool when the last slot was released, without checking the ownership of the process. That led to double-free or segfault if other processes/threads still referenced the pool (in shared memory scenarios).

After: The logic is updated to always check ownership (i.e. whether the pool exists in the local bucket) before releasing the pool. This also avoids removing an invalid pointer not tracked by the current process.

This patch does address the problem described in the PR.

catkit_core/HybridPoolAllocator.cpp
```diff
@@ -11,6 +15,29 @@ HybridPoolAllocator::HybridPoolAllocator(std::size_t capacity, std::size_t min_s
...
+				m_Allocator->Release(pool);
+			}
+			DEBUG_PRINT("Done with bucket.");
+		}
+		DEBUG_PRINT("Done with buckets for this thread.");
+	}
+	DEBUG_PRINT("Done with all buckets.");
+}
+
```
alexisyslau, line 40: _If I understand correctly:_
The destructor here is to ensure that pools will eventually be freed (by decrementing the ref_count by 1), even if all slots aren’t freed immediately. It is to prevent memory from being held due to the allocator’s ownership reference—being stuck at ref_count == 1 when all slots are released in a shared process?

_Question_
But out of curiosity, I don’t really know how it works here. I tried commenting the destructor out, but the pool still got removed without a problem. I did not run into errors even without the destructor. Of course, with the destructor, there’s no problem.

alexisyslau (approved):
Everything ran fine with the new test file — I didn’t run into any issues. I added a few comments with questions out of curiosity (still wrapping my head around some parts). But as far as I can tell, the accidental pool release issue with shared pools seems to be resolved under the current tests. Looks good to me!

catkit_core/HybridPoolAllocator.cpp
ehpor: No. While not specifically this line, but here is a more correct explanation.

Before: unconditionally _removes_ the pool, belonging to the slot that we are trying to deallocate, from the bucket. However, this pool might not have been allocated by us, so therefore that pool is not _in_ our bucket and therefore our process segfaults.

Intermediate: adding a check for that is easy. However, now if the process that allocated the pool (say process A) is not the one deallocating the last slot in that pool, the pool will be deallocated by process B, while process A still thinks it owns the pool. Therefore, on the next allocation, process A will try to use the freed pool. 

After: adding a check _and_ adding correct ref counting solves this intermediate problem. The allocator holds a reference to the pool. Furthermore, each allocated slot in the pool holds a reference to the pool as well. Therefore, the pool will be deallocated if and only if all slots are deallocated _and_ the allocator releases its reference. The latter happens in two cases: 1) the allocator itself is deallocated (see your comment below) or 2) the allocator that owns the pool deallocates the last slot.

catkit_core/HybridPoolAllocator.cpp
ehpor: The allocator still gets removed correctly, but it'll be leaking memory. Ie. some pools will not be deallocated and will live forever. So, if the destructor gets skipped too often (ie. due to a segfault somewhere else), then memory usage slowly accumulates. All this memory can be recovered by restarting the message broker (ie. the testbed server).

PR #342, opened 2025-06-12
Use interface for `MessageBroker`.

ehpor:
> [!CAUTION]
> This PR requires #331 to be merged before.

Since we're gonna have multiple message brokers, this PR creates a base class `MessageBroker` that contains the interface for all message brokers. All message brokers share the `Message` and `MessageSubscription` structs and their functionality.

This also renames the existing `MessageBroker` class into `LocalMessageBroker`.

FYI: this looks like a lot of code changes, but it's mainly moving code around rather than creating new code or modifying existing code.

### DIFF
--- a/catkit_core/TestbedProxy.cpp
+++ b/catkit_core/TestbedProxy.cpp
@@ -2,6 +2,7 @@
 
 #include "Timing.h"
 #include "HostName.h"
+#include "LocalMessageBroker.h"
 #include "testbed.pb.h"
 
 #include <memory>
@@ -382,7 +383,7 @@ void TestbedProxy::GetTestbedInfo()
 
 	m_MessageBrokerHeader = SharedMemory::Open(reply.message_broker_id());
 	StructStream stream = StructStream(m_MessageBrokerHeader);
-	m_MessageBroker = MessageBroker::Open(stream);
+	m_MessageBroker = LocalMessageBroker::Open(stream);
 
 	m_LoggingIngressPort = reply.logging_ingress_port();
 	m_LoggingEgressPort = reply.logging_egress_port();
--- a/tests/test_message_broker.py
+++ b/tests/test_message_broker.py
@@ -89,14 +89,14 @@ def test_message_broker_publish_array(broker):
     arr = np.random.randn(10)
     broker.publish_array('test_array', arr)
 
-    msg = broker.get_newest_message('test_array')
+    msg = broker.get_current_message('test_array')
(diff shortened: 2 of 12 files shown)


tests/test_message_broker.py
```diff
@@ -33,6 +30,9 @@ def test_message_subscription(broker):
     assert message_2 is not None
     assert message_3 is not None
 
+    assert message_1.frame_id == 2
```
steigersg: some inline comments would be helpful in this test/more descriptive naming of `message_1`, `message_2`, and `message_3` to explain the expected behavior. For example: 

`# message_1 will get the newest message (frame_id 2)`
`# message_2 will get the first sequential message (frame_id 0)`
`# message_3 will get the next sequential message (frame_id 1)`

catkit_core/TestbedProxy.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include "Timing.h"
 #include "HostName.h"
+#include "LocalMessageBroker.h"
```
steigersg, line 5: Is there a reason in the `TestbedProxy.h` file the `MessageBroker.h` header is included instead of the `LocalMessageBroker.h` header?

steigersg (dismissed):
Left a few comments, but due to the high priority of this PR do not think they warrant blocking it from being merged so I am approving it. Successfully migrates `MessageBroker` -> `LocalMessageBroker` while also clarifying some variable names and updating relevant tests.

catkit_core/TestbedProxy.cpp
ehpor: I do not want the things using a TestbedProxy to care what underlying implementation its message broker has. They should just be able to use it. However, in the source file, I need to create the object itself, which means that it needs access to the actual underlying object (which is then stored as a generic message broker for later use).

tests/test_message_broker.py
ehpor: Fixed in 19fb397.

ehpor:
I'm gonna do the rebase now. This will be rough.

ehpor:
Less painful than I thought. Only 7 merge conflicts, most of which were easy to fix.

@steigersg Reapproval is needed.

catkit_core/TestbedProxy.cpp
steigersg: thanks for clarifying!

steigersg (approved):
re-approving after rebase

PR #343, opened 2025-06-12
Add dark data stream to power meter

ivalaginja:
Adds a `dark` data stream to hardware and simulated power meter service, and updates the service doc.

Before merging:
Need to figure out how to not submit the default value 1. when trying to measure a dark in simulation. This is done on hardware by turning off the lights, which the power meter service has no knowledge of.

Might also consider removing the default zero frame on the stream since it would be better to get an error if no dark was measured yet instead of quietly just doing no dark correction (by subtracting zero).

### DIFF
--- a/catkit2/services/thorlabs_pm/thorlabs_pm.py
+++ b/catkit2/services/thorlabs_pm/thorlabs_pm.py
@@ -16,9 +16,7 @@ class ThorlabsPM(Service):
         self.interval = self.config.get('interval', 10)
 
         self.power = self.make_data_stream('power', 'float64', [1], 20)
-
         self.dark = self.make_data_stream('power', 'float64', [1], 20)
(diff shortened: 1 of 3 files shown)


ehpor:
Yep, don't submit the zero. No frame = no dark. 👍

ivalaginja:
Tested and works.

ivalaginja:
The route I decided to go was to blindly and continuously submit zeros to the dark data stream in simulation. In this way, if a non-zero number gets submitted to this stream in sim, it a newer frame containing zero again fairly soon. This is sufficient for the application in https://github.com/ivalaginja/thd-controls/pull/420, and only matters for simulated mode anyway.

ehpor (approved):
While I think that Thorlabs power meters have excellent dark correction on their own, I do agree that in the short-term this should be done.

PR #345, opened 2025-06-13
Enable Services to publish their CPU and memory usage.

ehpor:
This PR adds publishing of CPU usage and memory usage on the message broker. 

Tested on:
- [X] Windows
- [x] MacOS
- [x] Linux

### DIFF
--- a/catkit_core/ProcessStats.cpp
+++ b/catkit_core/ProcessStats.cpp
@@ -174,7 +174,7 @@ void ProcessStats::Update()
 	uint64_t current_task = getTaskTime();
 
 	if (current_total > m_LastTotalTime)
-		m_CpuPercent = 100.0 * (current_task - m_LastTaskTime) / (current_total - m_LastTotalTime);
+		m_CpuUsage = 100.0 * (current_task - m_LastTaskTime) / (current_total - m_LastTotalTime);
 
 	m_LastTotalTime = current_total;
 	m_LastTaskTime = current_task;
@@ -192,7 +192,7 @@ void ProcessStats::Update()
 	if (current_sys > m_LastSysJiffies)
 	{
 		int numCpus = sysconf(_SC_NPROCESSORS_ONLN);
-		m_CpuPercent = 100.0 * (current_proc - m_LastProcJiffies) / (current_sys - m_LastSysJiffies) * numCpus;
+		m_CpuUsage = 100.0 * (current_proc - m_LastProcJiffies) / (current_sys - m_LastSysJiffies) * numCpus;
 	}
 
 	m_LastProcJiffies = current_proc;
--- a/catkit_core/ProcessStats.cpp
+++ b/catkit_core/ProcessStats.cpp
@@ -1,0 +1,213 @@
+#include "ProcessStats.h"
+
+#include <iostream>
+#include <thread>
+
+#if defined(_WIN32)
+	#include <windows.h>
+	#include <psapi.h>
+#elif defined(__APPLE__)
+	#include <mach/mach.h>
+	#include <mach/host_info.h>
+	#include <mach/task_info.h>
+#else  // Linux
+	#include <fstream>
+	#include <sstream>
+	#include <unistd.h>
+#endif
+
+#if defined(_WIN32)
+
+static ULONGLONG fileTimeToUInt64(const FILETIME& ft)
+{
+	ULARGE_INTEGER li;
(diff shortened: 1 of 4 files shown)


alexisyslau:
I tested accessing the CPU and Memory usage by editing `tests/test_dummy_dm_commands.py`. Here's the output:

```
[testbed] Started service "dummy_dm_service" with type "dummy_dm_service".
[dummy_dm_service] Service was succesfully opened.
[dummy_dm_service] Starting service main function.
...
**CPU usage: 0.0**
**Memory usage: 152121344**
...
[testbed] Request received: shut_down
[dummy_dm_service] Server has shut down.
[dummy_dm_service] Service main has ended.
```
 Works for me (MacOS).

steigersg:
> I tested accessing the CPU and Memory usage by editing `tests/test_dummy_dm_commands.py`. Here's the output:
> 
> ```
> [testbed] Started service "dummy_dm_service" with type "dummy_dm_service".
> [dummy_dm_service] Service was succesfully opened.
> [dummy_dm_service] Starting service main function.
> ...
> **CPU usage: 0.0**
> **Memory usage: 152121344**
> ...
> [testbed] Request received: shut_down
> [dummy_dm_service] Server has shut down.
> [dummy_dm_service] Service main has ended.
> ```
> 
> Works for me (MacOS).

@alexisyslau I'm testing this on Linux and was wondering if you could share what tests you did on MacOS so that we're consistent?

alexisyslau:
> @alexisyslau I'm testing this on Linux and was wondering if you could share what tests you did on MacOS so that we're consistent?

I added these lines after line 17 in `tests/test_dummy_dm_commands.py` (_I think that’s the right way to access it?_):

```
# CPU usage
cpu_topic =  testbed.dummy_dm_service.id + "/cpu_usage/get"
cpu_msg   = testbed.message_broker.get_newest_message(cpu_topic)
if cpu_msg:
    cpu_usage = cpu_msg.payload[0]  
    print(f"CPU usage: {cpu_usage}")
else:
    print("No CPU usage message found.")

# Memory usage
mem_topic =
mem_msg   = testbed.message_broker.get_newest_message(mem_topic)
if mem_msg:
    memory_usage = mem_msg.payload[0] 
    print(f"Memory usage: {memory_usage}")
else:
    print("No memory usage message found.")
```

Then run pytest and make sure the prints are showing.

steigersg:
Thanks @alexisyslau! Just ran the same test on linux and got the following so it seems to be working and is consistent with your results. 
```
...
[testbed] Request received: start_service
[testbed] Trying to start service "dummy_dm_service".
[testbed] Starting new service with ['/home/hicattrois/miniconda3/envs/catkit2/bin/python', '/home/hicattrois/git_repos/catkit2/tests/services/dummy_dm_service/dummy_dm_service.py', '--id', 'dummy_dm_service', '--port', '44345', '--testbed_port', '6346'].
[testbed] Started service "dummy_dm_service" with type "dummy_dm_service".
...
**CPU usage: 0.0**
**Memory usage: 159207424**
...
[dummy_dm_service] Closing service.
[dummy_dm_service] Service was closed.
[testbed] Service "safety" appears to have crashed.
[testbed] Server has shut down.
```
Note there was a typo in the test posted above as it should be 
`mem_topic = testbed.dummy_dm_service.id + "/memory_usage/get"`

Also I believe the test script is called `test_dm_commands.py` not `test_dummy_dm_commands.py`

steigersg:
@ehpor could we add that test (or similar) to `test_message_broker.py` before merging?

ehpor:
You want me to test that those two topics are submitted? Or that the values are correct?

Also, this is not part of the message broker, it's part of the service code.

steigersg:
> You want me to test that those two topics are submitted? Or that the values are correct?
> 
> Also, this is not part of the message broker, it's part of the service code.

I was just thinking that they're submitted and can be accessed. If you don't think its necessary that's fine too, just the fact that this needed to be tested on multiple OS's indicated to me that it could be useful to add a test in case this code is updated in the future and we're in this same position.

alexisyslau (approved):
Tested it out on Linux, macOS (arm, not intels, not sure if they change anything between versions), and Windows. The service grabs CPU and memory usage just fine, and both are available via the message broker. As for the test part, I doubt they'd change the standard headers across OSes — but if that ever happens, it should throw a compile-time error anyway.

PR #346, opened 2025-06-14
Fix memory lifetime for Shareable objects.

### LINKED ISSUE #344: Shareable objects rely on the memory staying allocated, but this is not enforced.
It would be nice to have the memory be reference counted, but I don't know if this is possible to do with the current infrastructure of `StructStream`, since that object takes a raw memory pointer.

Maybe it could take a `Memory` object with offset and construct a stream from that? The `Memory` object would be reference counted. Then, the `Shareable` base class could take a reference to that memory object upon construction, ensuring that the memory stays active until the `Shareable` is deallocated.

FYI: this bug has cost me ~~4~~ ~~6~~ 7.5 hours so far. (Future Emiel: update this number when you lose more time.)

ehpor:
Fixes #344.

This PR makes one important change. When a `Shareable` object is now created, you can pass in a shared pointer to a `Memory` object. This object is stored in the object itself, thereby keeping that memory alive until the `Shareable` object has been deallocated. This means that there will be no more cases where the memory got deallocated while the `Shareable` is still alive (which almost always leads to access errors or segmentation faults).

While this PR touches a lot of code, almost all code changes are similar: passing the buffer that the `StructStream` is based on into the constructor of `Shareable`.

Additionally, this PR splits up the header-only `StructStream.h` into a `StructStream.h`, `StructStream.cpp` and `StructStream.inl` files, as is more common in our codebase.

### DIFF
--- a/benchmarks/buddy_allocator.cpp
+++ b/benchmarks/buddy_allocator.cpp
@@ -1,5 +1,7 @@
 #include "BuddyAllocator.h"
 #include "Timing.h"
+#include "LocalMemory.h"
+
 #include <iostream>
 
 void benchmark_linux_scalability()
@@ -11,7 +13,7 @@ void benchmark_linux_scalability()
 	auto *handles = new BuddyAllocator::Handle[N];
 
 	size_t buffer_size = BuddyAllocator::GetSharedStateSize(MAX_SIZE, BLOCK_SIZE);
-	char *buffer = new char[buffer_size];
+	auto buffer = LocalMemory::Create(buffer_size);
 
 	auto stream = StructStream(buffer);
 	auto allocator = BuddyAllocator::Create(stream, MAX_SIZE, BLOCK_SIZE);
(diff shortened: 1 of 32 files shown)


alexisyslau (dismissed):
This is a solid refactor that prevents access violations and segmentation faults that could occur when a `Shareable` object remains alive but the memory it operates on has already been deallocated — for example, if the object is still referencing memory that no longer exists. Additionally, it removes the need for manual memory management.

One quick note for Emiel: some changes have already been made, but parts of the benchmark code may still need to be updated or refactored to align with the new definitions.

ehpor:
@alexisyslau 

> One quick note for Emiel: some changes have already been made, but parts of the benchmark code may still need to be updated or refactored to align with the new definitions.

What do you mean by this? I didn't run the benchmarks again, but the code compiled just fine so it can't be all wrong.

alexisyslau:
> @alexisyslau
> 
> > One quick note for Emiel: some changes have already been made, but parts of the benchmark code may still need to be updated or refactored to align with the new definitions.
> 
> What do you mean by this? I didn't run the benchmarks again, but the code compiled just fine so it can't be all wrong.

Ah, sorry, there was a typo. I meant the codes under` benchmarks/`. Just thinking that if any of those tests manually allocate memory to exercise `Shareable` behavior, they might need to be tweaked a bit. But you're right, if it compiles cleanly, things are fine overall.

ehpor:
@alexisyslau

Yes, all those benchmarks use `Shareable` objects in some way. I think I fixed all of the issues, but indeed I didn't run the code. I'll fix any issues later in future PRs, when I need to rerun benchmarks.

I need a re-approval after the rebase due to fixing of a merge conflict.

alexisyslau (approved):
Looks good to me after the rebase!

PR #350, opened 2025-06-19
Add tests for process stats.

ehpor:
> [!CAUTION]
> This PR branches off from #342.

In response to https://github.com/spacetelescope/catkit2/pull/345#issuecomment-2985134924.

This PR adds a wrapper for the `ProcessStats` class and then test the output using pytest. This ensures that the functions do not crash and give at least a minimal reasonable output (memory usage positive, cpu usage non-negative).

### DIFF
--- a/catkit2/bindings.cpp
+++ b/catkit2/bindings.cpp
@@ -36,6 +36,7 @@
 #include "Event.h"
 #include "Uuid.h"
 #include "ArrayView.h"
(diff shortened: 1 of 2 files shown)


steigersg (dismissed):
Looks good to me! (only looking at the changes not in #342) - thanks!

steigersg (approved):
re-approving after rebase

PR #351, opened 2025-06-20
Fix mtop call for `get_newest_message()`.

ehpor:
This function was renamed to `get_current_message()`.

If the function fails, that means that there was no message on this topic, we we continue to the next topic.


ehpor:
Superseded by #353 and therefore no longer necessary.

PR #353, opened 2025-06-25
Use textual instead of curses for mtop.

ehpor:
In response to @lanemeier7 finding that `mtop` required `curses` or `curses-windows` to be installed, I switched `mtop` to use textual, a library for making terminal apps. This makes `mtop` slightly easier to use, themeable with an actual help page, but most of all, easier to develop further.

## Screenshots
image

image

image

### DIFF
--- a/catkit2/tui/mtop.py
+++ b/catkit2/tui/mtop.py
@@ -1,5 +1,10 @@
-import curses
-import time
+from textual.app import App, ComposeResult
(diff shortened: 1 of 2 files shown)


lanemeier7:
I just tested this out on my Windows machine and this works great!

lanemeier7 (approved):
This works for me and I approve!

PR #355, opened 2025-06-25
ZWO Cool Camera Support

### LINKED ISSUE #336: ZWO 174mm Cool Support
Problem
When attempting to initialize a service with the ZWO 174mm cool camera, the service fails with the following error:

```
Critical Error: Shutting down service.
```
Root Cause
The initialization process is attempting to set the CoolPowerPerc control value, which appears to be causing the service to crash.

Proposed Solution
Skip setting the CoolPowerPerc control value during service initialization by adding a conditional check:
``` python
if c != 'CoolPowerPerc':
    self.camera.set_control_value(controls[c]['ControlType'], controls[c]['DefaultValue'])
```

lanemeier7:
Adding ZWO Cool camera support to catkit2 and closes #336 

The ZWO Cool camera has some new controls with regards to temperature setting and monitoring. Catkit2 does not work out of the box with these as it tries to program fields that are not writable. 

Additionally our ZWO cameras occasionally struggles to take large images when the bandwidth is set to the max value which is the catkit2 default and this field is necessary for the high data rates which we like to run the zwo at. 

In order to get around this I have added a max_bandwidth field which when True (default) uses max bandwidth and the default catkit behavior, however when false it sets the minimum bandwidth and this can help take images when the camera or cabling is in certain states. Perhaps USB2 is plugged in and not USB3, and also we have replugged cables and seen better / different behavior.

### DIFF
--- a/catkit2/services/zwo_camera/zwo_camera.py
+++ b/catkit2/services/zwo_camera/zwo_camera.py
@@ -95,14 +95,21 @@ class ZwoCamera(Service):
 
         # Restore all controls to default values, in case any other application modified them.
         for c in controls:
-            self.camera.set_control_value(controls[c]['ControlType'], controls[c]['DefaultValue'])
+            # Some control values such as Temperature and CoolerPowerPerc are not writable for zwo cool models
+            if ('IsWritable' in controls[c]) and controls[c]['IsWritable']:
+                self.camera.set_control_value(controls[c]['ControlType'], controls[c]['DefaultValue'])
 
         print('Bandwidth defaults', self.camera.get_controls()['BandWidth'])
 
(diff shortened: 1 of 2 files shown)


ehpor (dismissed):
I don't have a ZWO camera to test this with, but the code looks fine to me.

lanemeier7:
I want to test this on one of our non-cool model ZWO cameras first just to be sure I don't break anything on HiCAT. This has been tested on the ZWO 174mm cool camera

lanemeier7:
@ehpor I made a minor change to the documentation to explain the new max_bandwidth parameter. I also tested this out and verified everything is still working on our ZWO 1600mm camera (and yes it all still works).

ehpor (approved):

PR #360, opened 2025-07-03
Hamamatsu camera temperature cooling

Corentin-Pa:
Put the water cooling and the fan to the mode defined in the `config.yml` file.

### DIFF
--- a/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
+++ b/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
@@ -101,11 +101,7 @@ class HamamatsuCamera(Service):
         self.cooling_mode = self.config.get('cooling_mode', 'on')
         self.fan_on = self.config.get('fan_on', True)
 
-        if self.fan_on:
-            fan = 1
-        else:
-            fan = 0
-
+        fan = 1 if self.fan_on else 0
         self.cam.prop_setvalue(dcam.DCAM_IDPROP.SENSORCOOLERFAN, fan)
 
         if self.cooling_mode == 'off':
--- a/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
+++ b/catkit2/services/hamamatsu_camera/hamamatsu_camera.py
@@ -97,6 +97,9 @@ class HamamatsuCamera(Service):
         if self.cam.dev_open() is False:
             raise RuntimeError(f'Dcam.dev_open() fails with error {self.cam.lasterr()}')
 
+        # Set temperature cooling to "ON" (tarder =-20)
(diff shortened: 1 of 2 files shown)


ivalaginja (changes requested):
We don't want the input values of the camera setters to be hard-coded. Somebody else might want to be sue the camera with the fan on for example.

The way to fix this is to read the values you want to set from the configuration. This also means that you will have to open a pair-PR over on thd-controls in which you add the values we need to the camera section of the configuration file.

We probably also want to be able to use values that are a little more descriptive than random floats. So I think the config file should carry a key like `cooling_mode` which can be set either to `on`, `off` or `max`. Inside the service, you then convert those strings to the appropriate floats used in the actual function.

Similar, a key called `fan_on` should take in `true` or `false`, which can then be converted to `0` if it's on and so forth.

Also remember to Adapt the comments accordingly.

catkit2/services/hamamatsu_camera/hamamatsu_camera.py
```diff
@@ -1,513 +1,535 @@
...
+
+        # Set water cooling mode and fan mode
+        self.cooling_mode = self.config.get('cooling_mode', 'on')
+        self.fan_on = self.config.get('fan_on', True)
+
+        if self.fan_on:
+            fan = 1
+        else:
+            fan = 0
+
```
ivalaginja: If you want to be fancy, you can simplify this to: `fan = 1 if self.fan_on else 0`

ivalaginja (approved):
Very nice!

ivalaginja:
I unapproved because I remembered we need a check on hardware first x)

ivalaginja (changes requested):
HW test needed

Corentin-Pa:
Tested on hardware with the following test : In a ipython terminal, check the fan and the cooling status are respectively on and off according to the `config.yml` file, then change the values in the  `config.yml` and recheck in the ipython terminal the fan and cooling status.

Working as expected.

ivalaginja (commented):
Looking good. Could you please add the two new config parameters to the Hamamatsu service docs as well? you can find them here:
https://github.com/spacetelescope/catkit2/blob/develop/docs/services/hamamatsu_camera.rst

No retesting will be needed after that change.

Corentin-Pa:
I have already put them in the configuration section of the doc. Do you want it somewhere else too ?

ivalaginja:
> I have already put them in the configuration section of the doc. Do you want it somewhere else too ?

Oh I got confused between the real configuration file on thd-controls and the docs on catkit2, my bad!

Nothing else needed, good to go!

ivalaginja (approved):

PR #361, opened 2025-07-03
Add power function mode to Thorlabs CLD laser diodes

ivalaginja:
Previously, we could only use this service to control a CLD by controlling the LD current. This PR adds the option to control the the LD power by controlling the photodiode current instead.

Software docs:
https://www.thorlabs.com/drawings/3d9d377a333d5fcc-242A96AC-CDBB-1019-3464EF2B8B26961B/CLD1015-ProgrammersReferenceManual.pdf

And specifically:
Screenshot 2025-07-03 at 19 35 06


ivalaginja:
Tested on hardware and this works.

a-sevin (approved):

PR #366, opened 2025-07-11
Error in parameter name Hamamatsu

johanmazoyer:
Difference between name of param in the service and in the aprameter file for defect_correction parameter. 
Chose the param file name as it makes more sense.

Old branch that was never PRed


ivalaginja (approved):
LGTM

PR #367, opened 2025-07-16
Add Thorlabs rotation stages to Kinesis motor controls

Lukasdly:
Add a new motor in the list into the thorlabs_cube_motor_kinesis.py file.
Complementary work for the outcome : https://github.com/ivalaginja/thd-controls/issues/454. 
Support https://github.com/ivalaginja/thd-controls/pull/456

### DIFF
--- a/catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
+++ b/catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
@@ -109,9 +109,9 @@ class ThorlabsCubeMotorKinesis(Service):
             gear_box_ratio = c_double(67.49)
             pitch_mm = c_double(1.0)
         elif self.stage_model in['PRM1Z8'] :
-            steps_per_rev = c_double(512) # Same as other motors
-            gear_box_ratio = c_double(67) # Want to fit the value in the manual of PRM1Z8
-            pitch_mm = c_double(1.0) # Don't know if it make sens with mm : we're with degree...
+            steps_per_rev = c_double(512)  # Same as other motors
+            gear_box_ratio = c_double(67)  # Want to fit the value in the manual of PRM1Z8
+            pitch_mm = c_double(1.0)  # Don't know if it make sens with mm : we're with degree...
         else:
             raise ValueError(f"Stage model {self.stage_model} not supported.")
 
--- a/catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
+++ b/catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
@@ -71,7 +71,6 @@ class ThorlabsCubeMotorKinesis(Service):
         self.lib: CDLL = cdll.LoadLibrary(default_dll_path + dll_name)
 
         self.unit = None
-        self.unit_type = 0
         self.min_position_device = None
         self.max_position_device = None
         self.min_position_config = None
@@ -108,10 +107,12 @@ class ThorlabsCubeMotorKinesis(Service):
             steps_per_rev = c_double(512)
             gear_box_ratio = c_double(67.49)
             pitch_mm = c_double(1.0)
+            self.unit = 'mm'
         elif self.stage_model in ['PRM1Z8']:
-            steps_per_rev = c_double(512)  # Same as other motors
-            gear_box_ratio = c_double(67)  # Want to fit the value in the manual of PRM1Z8
+            steps_per_rev = c_double(1919.64186)  # Same as other motors
+            gear_box_ratio = c_double(1.0)  # Want to fit the value in the manual of PRM1Z8
             pitch_mm = c_double(1.0)  # Don't know if it make sens with mm : we're with degree...
+            self.unit = 'degree'
         else:
             raise ValueError(f"Stage model {self.stage_model} not supported.")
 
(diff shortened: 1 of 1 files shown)


Lukasdly:
I need to discuss the value I take: I'm not sure about it.

ivalaginja:
> I need to discuss the value I take: I'm not sure about it.

@Corentin-Pa would you have an idea where to find the correct values? @Lukasdly you could probably go to the Thorlabs website, find the new motor stage you are adding on there and download the manual/spec sheet for it. There should be information like this inside there.

Lukasdly:
@ivalaginja I already look into the manual but it does not correspond to values you put with other motor (67 =! 67.49)
 steps_per_rev = c_double(512)  : **I think i don't have to change it**
  gear_box_ratio = c_double(67)  : **If I understand correctly, this value will ensure the correct conversion between degrees and motor steps. It may need to be changed**
  pitch_mm = c_double(1.0) :  **Don't know if it make sens with mm : we're with degree...**

ivalaginja:
> @ivalaginja I already look into the manual but it does not correspond to values you put with other motor (67 =! 67.49) steps_per_rev = c_double(512) : **I think i don't have to change it** gear_box_ratio = c_double(67) : **If I understand correctly, this value will ensure the correct conversion between degrees and motor steps. It may need to be changed** pitch_mm = c_double(1.0) : **Don't know if it make sens with mm : we're with degree...**

@Lukasdly What exactly do you mean with "the manual"?

Lukasdly:
> > @ivalaginja I already look into the manual but it does not correspond to values you put with other motor (67 =! 67.49) steps_per_rev = c_double(512) : **I think i don't have to change it** gear_box_ratio = c_double(67) : **If I understand correctly, this value will ensure the correct conversion between degrees and motor steps. It may need to be changed** pitch_mm = c_double(1.0) : **Don't know if it make sens with mm : we're with degree...**
> 
> @Lukasdly What exactly do you mean with "the manual"?

The manual of the motor : https://www.thorlabs.com/thorproduct.cfm?partnumber=PRM1/MZ8#ad-image-0

ivalaginja:
I don't even know how, but I found this:
https://github.com/Thorlabs/Motion_Control_Examples/blob/c73742dcfdbbd61e823ea0c38946b7a3e98469ac/Python/KCube/KDC101/kdc101_example.py#L49C9-L49C22

The question is, what to do with it.

Corentin-Pa:
The manual for the motor is really not clear. I would put the encoder count per mm at 1919.6418578623391 as the manual here shows (just search for the PRM1-Z8 in it) : https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf

For the other values I would put it at 1 for now. I think `1919.6418578623391` is directly the scaling factor between the encoder counts and a whole rotation of 360°. It is not the same thing as the `512` we have for the other motors, which is the encoder counts per revolution of the motor. The encoder counts for 1mm was converted probably elsewhere with the gearbox ratio and the pitch. 

But if we put a gear and a pitch of 1 I think it would probably work. However I don't know what the line `self.lib.CC_SetMotorParamsExt(self.serial_number, steps_per_rev, gear_box_ratio, pitch_mm)` does exactly so it's hard to tell.

So to sum up, for now I would just do this and see what it does : 
```
        elif self.stage_model in ['PRM1Z8']:
            steps_per_rev = c_double(1919.6418578623391) 
            gear_box_ratio = c_double(1.0)
            pitch_mm = c_double(1.0)
```

ivalaginja:
This coincides with the code example I found. @Lukasdly seems like you got your answer?

THD2-team:
> The manual for the motor is really not clear. I would put the encoder count per mm at 1919.6418578623391 as the manual here shows (just search for the PRM1-Z8 in it) : https://www.thorlabs.com/Software/Motion%20Control/APT_Communications_Protocol.pdf
> 
> For the other values I would put it at 1 for now. I think `1919.6418578623391` is directly the scaling factor between the encoder counts and a whole rotation of 360°. It is not the same thing as the `512` we have for the other motors, which is the encoder counts per revolution of the motor. The encoder counts for 1mm was converted probably elsewhere with the gearbox ratio and the pitch.
> 
> But if we put a gear and a pitch of 1 I think it would probably work. However I don't know what the line `self.lib.CC_SetMotorParamsExt(self.serial_number, steps_per_rev, gear_box_ratio, pitch_mm)` does exactly so it's hard to tell.
> 
> So to sum up, for now I would just do this and see what it does :
> 
> ```
>         elif self.stage_model in ['PRM1Z8']:
>             steps_per_rev = c_double(1919.6418578623391) 
>             gear_box_ratio = c_double(1.0)
>             pitch_mm = c_double(1.0)
> ```

Here Lukas at CR ; we found the same values with Pierre and it works in the lab. @ivalaginja @Corentin-Pa

Lukasdly:
The test was to run the new example.py 

> Homing work well

> Motor can rotate in two directions and it worked

> Rotation worked as an absolute rotation or as a relative rotation

> We check that the values in degree in the console are the same that the values we see directy on the polarizer

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
```diff
@@ -108,6 +107,12 @@ def open(self):
             steps_per_rev = c_double(512)
             gear_box_ratio = c_double(67.49)
             pitch_mm = c_double(1.0)
+            self.unit = 'mm'
+        elif self.stage_model in ['PRM1Z8']:
```
ivalaginja: Let's try to keep consistency and call this `PRM1-Z8` with the dash, like the other motors.

ivalaginja (commented):
Just one minor request, no retesting needed when you fix it.

Also no retesting needed if you rebase since the PR that went in on the other side jsut touched the Orca.

Lukasdly:
I change the name in the thorlabs_cube_motor_kinesis.py and also in services.yml into THD-control. I also manage comments.

ivalaginja (commented):
Good to go

ivalaginja (approved):

catkit2/services/thorlabs_cube_motor_kinesis/thorlabs_cube_motor_kinesis.py
```diff
@@ -131,13 +136,6 @@ def open(self):
         self.min_position_device = min_position.value
         self.max_position_device = max_position.value
 
-        # Get the unit of the motor (mm or deg).
-        self.unit_type = self.lib.CC_GetMotorTravelMode(self.serial_number)
-        if self.unit_type == 1:
-            self.unit = 'mm'
-        elif self.unit_type == 2:
-            self.unit = 'deg'
```
Lukasdly, line 139: `CC_GetMotorTravelMode()` is a broken function: it always returns 1 . So we changed it.

PR #364, opened 2025-07-08
Hamamatsu temperature safety

johanmazoyer:
Added setter and getter for 
- fan status (on/off)
- Cooler mode (off/on/max)

Added safety based on temperature. 
If temperature > 28 degrees (can be changed in yaml parameter), fan start and acquisition stops.
Fix issue: https://github.com/ivalaginja/thd-controls/issues/419 

The following properties (documented in the support parameter manual and/or directly in the dcamapi4.py python file) do not seem to work, at least not for orca quest 2 (C15550-22UP): 
- SENSORCOOLERSTATUS supposed to return "busy" when camera is still cooling to target
- SENSORTEMPERATURE_STATUS supposed to return a warning in case of error in the temperature ? (not documented in the manual, only appear in dcamapi4.py)

PR #368, opened 2025-07-16
Use Enums for cooler and fan settings

ivalaginja:
Suggested change for #364, which is otherwise ready. Enums are a much safer way to map values, and it also removes the requirement to come up with random function names.

If you could give this a spin on hardware @johanmazoyer then we could merge the two PRs sequentially.


johanmazoyer:
Updated change based on your modifs and my comments :
- fan_status is now a string (on/off) instead of bool, to simplify conversion and be similar to Cool_mode
- simplified enum by using existing functionality 

Need to be tested on testbed

ivalaginja:
So much better!

johanmazoyer:
Working on testbed !

PR #369, opened 2025-07-17
Added subframe function for FLIR service

sfr348:
Added subframe function to FLIR service to enable faster readout. Tested on Ubuntu. Written by @apearsonn. Screenshots coming soon.


sfr348:
Will also fix flake8 things before requesting review.

ehpor:
And this isn't a function on the proxy?

sfr348:
I don't see it anywhere?

ehpor:
I mean: can't you do this as a function on the proxy? This feels like a convenience function to me.

PR #373, opened 2025-07-22
Separate NKT SuperK services for EVO and FIANIUM

### LINKED ISSUE #372: Separate service classes for NKT Evo and NKT Fianium
They're not fully compatible even if they share the Varian when used, which can be copy-pasted.

Main concerns that support two separate classes are different class attributes (hardware getters and setters) and different safety measures for not burning fibers (on the Fianium managed with the pulse picker ratio which the Evo does not have).

Preliminary todo list for this issue:

- [ ] Revert current class to what it was and rename to identify it for the Evo
- [ ] Copy-paste that into a new Fianium class
- [ ] Ensure that the `set_spectrum()` inside the Fianium uses the pulse picker ratio safety that is wrapped around the pulse picker ratio data stream.
- [ ] Adapt configs of THD2 and HiCAT

There's probably more that I'm not thinking of right now.

ivalaginja:
Fixes #372.

@ehpor did you have it in mind like this?

### TODO
- [x] Address question about pulse picker ratio in `close()`
- [x] Finish logic around pulse picker ration in proxy function `set_spectrum()`
- [x] Test on hardware on THD2.
- [x] Update services.yml on HiCAT for compatibility -> https://github.com/spacetelescope/hicat-package2/pull/720


ivalaginja:
Works on hardware as expected, able to set central wavelengths and bandwidths through the proxy function, and power and pulse picker ratio through data streams.

ehpor (approved):
Merging this now. Didn't see anything egregious. While this is not how we want it, schedule wise we're going to merge this now and deal with doing a proper base class later

PR #374, opened 2025-07-24
Add proxy for optical fiber switch

ivalaginja:
Needs test on hardware.


catkit2/testbed/proxies/optical_fiber_switch.py
```diff
@@ -0,0 +1,32 @@
...
+            # The channel may still be a named channel, so try to resolve deeper.
+            return self.resolve_channel(channel)
+        else:
+            return channel
+
+    @property
+    def channels(self):
+        return self.config['channels']
+
+    def get_named_channel(self, channel):
```
ehpor, line 30: You won't find some recursive channel names this way. As in, a=1, b=2, nominal=a, then 1 won't give back nominal as a channel. You're probably not using this in that way, but it's something to be aware of.

ehpor (approved):

PR #380, opened 2025-08-03
Add default input channel for simulated optical switch

ivalaginja:
The data stream that is supposed to provide the input channel is empty upon creation of the simulated service for the optical switch. I made it contain 1 as a default when the service opens.

Code sneak: returned named channel from the proxy should not be a list.


ivalaginja:
@Corentin-Pa @ehpor whoever of you gets to this first, feel free to merge immediately if you see no problems with this PR.

Corentin-Pa (approved):
Good to go!

PR #382, opened 2025-08-04
Treat pulse picker ratio in sim service exactly like in hw service

ivalaginja:
Which is to give it a safe value of 100 upon startup and upon closing of the service.

Tested and works.


Corentin-Pa (approved):
Good to go!

PR #385, opened 2025-09-05
Fix data stream population in optical fiber switch service

ivalaginja:
The data stream `current_position` was not being populated correctly when the input channel was changed. This PR fixes this.

Tested on hardware and works now as expected.


Corentin-Pa (approved):
Good to go

PR #387, opened 2025-09-29
Add testbed minimal working example

ivalaginja:
Code originally written by @erinpougheon.

I call it "Benchlet", let me know if people prefer something else.

Todo:
- [x] Check all imports
- [x] Add entry points for Benchlet CLI (see comment below)
- [x] Test

Note: This PR or #389 should be merged but not both


ivalaginja:
This needs to go somewhere:

```python
from setuptools import setup, find_packages

setup(name="benchlet",
      version="0.1.0",
      author="The CATKit2 team",
      description="Control code and simulator for an example catkit2 testbed.",
      url="https://github.com/spacetelescope/catkit2",
      packages=find_packages(),
      package_data={'benchlet': ['user_interface/assets/*']},
      classifiers=[
                        "Programming Language :: Python :: 3"
                  ],
      python_requires='>=3.6',
      zip_safe=False,
      install_requires=[],
      entry_points={
            "console_scripts": [
                  "benchlet=benchlet.cli_interface:main"
            ],
      },
      )
```

ivalaginja:
@steigersg @RemiSoummer passing on to you guys. Might want to let the THD2 folks know when this is merged.

steigersg:
@RemiSoummer "benchlet" is now working and tested (in simulation). There are a couple of higher level questions I have/decisions that need to be made however before this is ready for review

1. Do we like "benchlet" as a name? I think its cute but maybe not the most descriptive.
2. Do we want benchlet to be its own package within the catkit2 repo (my preference), or a subpackage of catkit2? We can do either one and neither option is more complicated.  The main difference is in the first scenario (benchlet is its own package) the user would have to do a separate pip install of benchlet. The pro is that this keeps it separate from catkit2 and so is more accurate to what a users own testbed repo that uses catkit2 would look like.

RemiSoummer:
Thanks @ivalaginja and @steigersg  
I am going to close this one, as this is superseded by cookie cutter #389 
having the template makes it even easier and more flexible.

PR #388, opened 2025-10-16
Add cross-platform SIGINT handling in testbed.py

a-sevin:
Implement cross-platform interrupt handling for processes.


ehpor (changes requested):
Nice work!

I disagree with the fallback. We make the distinction between interrupting and terminating the service. Interrupting means that the service will still shut down cleanly and is used in cases where somehow it's not responding to requests over TCP to shut itself down. The SIGINT is captured by Python, results in a KeyboardInterrupt exception which can dislodge stubborn processes. On the other hand, termination is unclean and can therefore result in the process leaving a device inoperable until the computer is restarted (in the worst case). Therefore, if I issue an "interrupt", I do not expect the process to be killed uncleanly.

a-sevin:
> I disagree with the fallback. We make the distinction between interrupting and terminating the service. Interrupting means that the service will still shut down cleanly and is used in cases where somehow it's not responding to requests over TCP to shut itself down. The SIGINT is captured by Python, results in a KeyboardInterrupt exception which can dislodge stubborn processes. On the other hand, termination is unclean and can therefore result in the process leaving a device inoperable until the computer is restarted (in the worst case). Therefore, if I issue an "interrupt", I do not expect the process to be killed uncleanly.

OK, I think I understand the issue but how to do plan to handle the fallback correctly. If you issue an SIGINT and the process is not responding. What do you do ?

ehpor:
Either the user was the one issuing the interrupt and in that case, they'd issue a terminate afterwards when the service didn't shut itself down. Or in the case of the server shutting down, it's again the user that decides to terminate rather than interrupt.

Am I mistaken in this reasoning?

a-sevin:
No, that's fine, I didn't understand.

The idea is to remove the `terminate` and just log the error.

In the code, I replace this part

```python
        except (OSError, psutil.NoSuchProcess, PermissionError):
            # If anything goes wrong, fall back to terminating the process
            # to avoid leaving it stuck. The caller can choose to escalate.
            try:
                if self.process:
                    self.process.terminate()
            except (OSError, psutil.NoSuchProcess, PermissionError) as term_exc:
                logger = globals().get("logger", None)
                if logger:
                    logger.warning(f"Failed to terminate process {self.process.pid if self.process else 'unknown'}: {term_exc}")
                else:
                    print(f"Warning: Failed to terminate process {self.process.pid if self.process else 'unknown'}: {term_exc}", file=sys.stderr)
```

by

```python
        except (OSError, psutil.NoSuchProcess, PermissionError) as term_exc:
            logger = globals().get("logger", None)
            if logger:
                logger.warning(f"Failed to interrupt process {self.process.pid if self.process else 'unknown'}: {term_exc}")
            else:
                print(f"Warning: Failed to interrupt process {self.process.pid if self.process else 'unknown'}: {term_exc}", file=sys.stderr)
```

a-sevin:
@ehpor , are you ok with the modification ?

ehpor (approved):
Sorry for the delay. This slipped through my notifications.

PR #390, opened 2025-11-13
Miniforge Service Cleanup Bugfix

### LINKED ISSUE #359: Services crashing on shut down with Python 3.8.20, Windows & Miniforge
The ExoSpec lab is running Python 3.8.20 and all packages installed via miniforge on Windows 10/11.

In this environment, shutting down services causes Python processes to crash with pop up message "Python has stopped working" in the ~ mid June version of develop.

Image

 This message hangs until the user clicks "close".

lanemeier7:
### Problem
On Windows 10/11 with Python 3.8.20 (miniforge), shutting down catkit2 services causes Python processes to crash with a Windows system popup: "Python has stopped working". This popup hangs the application until the user manually clicks "Close".

image

This issue was introduced in the mid-June version of the `develop` branch.

### Root Cause
The explicit call to `CleanupAttributes()` at the end of `Service::Run()` was causing premature cleanup of `m_Properties`, `m_Commands`, and `m_DataStreams` containers. These containers hold shared pointers to objects that may still have active dependencies or references at that point in the shutdown sequence. On Windows, this premature cleanup triggers an access violation that manifests as a structured exception (SEH), which cannot be caught by standard C++ try-catch blocks.

### Solution
Removed the explicit `CleanupAttributes()` call from `Service::Run()`. The containers are now cleaned up naturally when the Service destructor runs, ensuring proper destruction order based on member variable declaration order in the class. This approach is safer and follows C++ best practices for RAII (Resource Acquisition Is Initialization).

### Changes
- Removed call to `CleanupAttributes()` in `Service::Run()`
- Added comment explaining the rationale
- Retained the `CleanupAttributes()` method implementation for potential future use

### Testing
- Verified services shut down cleanly on Windows 10/11 without crash popups
- Environment: Python 3.8.20, miniforge, Windows 10/11

### DIFF
--- a/catkit_core/Service.cpp
+++ b/catkit_core/Service.cpp
@@ -237,7 +237,7 @@ void Service::Run(void (*error_check)())
 	std::uint64_t timestamp = 0;
 	m_Heartbeat->SubmitData(&timestamp);
 
-		ArrayInfo info{'u', '=', 8, 1, {1, 1, 1, 1}, {8, 1, 1, 1}};
+	ArrayInfo info{'u', '=', 8, 1, {1, 1, 1, 1}, {8, 1, 1, 1}};
 	m_Testbed->GetMessageBroker()->PublishArray(m_ServiceId + "/heartbeat/get", {info, &timestamp});
 
 	// CleanupAttributes() removed - let natural destruction handle cleanup
--- a/catkit_core/Service.cpp
+++ b/catkit_core/Service.cpp
@@ -28,7 +28,7 @@ Service::Service(string service_type, string service_id, int service_port, int t
 	: m_Server(service_port), m_ServiceId(service_id), m_ServiceType(service_type),
 	m_LoggerConsole(), m_LoggerPublish(),
 	m_Heartbeat(nullptr), m_State(nullptr), m_Safety(nullptr), m_Testbed(nullptr),
-	m_IsRunning(false), m_ShouldShutDown(false), m_FailSafe(false)
+	m_IsRunning(false), m_ShouldShutDown(false), m_FailSafe(false), m_IsBeingDestroyed(false)
 {
 	m_Testbed = make_shared<TestbedProxy>("127.0.0.1", testbed_port);
 	m_Config = m_Testbed->GetConfig()["services"][service_id];
@@ -62,6 +62,12 @@ Service::Service(string service_type, string service_id, int service_port, int t
 
 Service::~Service()
 {
+	// Mark that we're being destroyed to prevent any attribute access during cleanup
+	m_IsBeingDestroyed = true;
+
+	// Explicitly clean up attributes in a safe order
+	// This prevents circular references from keeping the Service alive
+	CleanupAttributes();
 }
 
 void Service::Run(void (*error_check)())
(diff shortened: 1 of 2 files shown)


ehpor:
I totally agree that the `CleanupAttributes()` function was basically a kludge. But it solved an invisible problem: attributes contain a reference to the service object. This means that when the service object falls out of scope in `main()`, that the reference count will not become zero (due to circular references) and that the service object will not be cleaned up at all. In some cases that leads to shared memory not being cleaned up properly. 

I agree that this needs to be fixed, but at that moment, I didn't have the time for more than a messy workaround. Maybe attributes should have a weak reference instead of a strong one?

ehpor:
Secondary question: how is anything still using the properties/datastreams/commands at that point?

ehpor:
Nope. The destructor won't get called so moving the cleanup in the destructor won't change anything. The destructor only gets called when the object itself gets removed, not when the shared_ptr gets removed.

lanemeier7:
Ok - here is another stab at it. This seems to work on my machine but would want to make sure it doesn't destroy the other environments.

### Root Cause
The service architecture has two interconnected issues:

1. **Circular References**: Properties and Commands capture `this` in their lambda functions, creating strong references back to the Service object. When the Service object goes out of scope, these circular references prevent proper cleanup, leading to memory leaks and shared memory not being freed.

2. **Unsafe Cleanup Timing**: The original code called `CleanupAttributes()` at the end of `Service::Run()`, attempting to break circular references. However, on Windows this caused access violations when clearing the containers while the Service object was still accessible from other threads or Python code, resulting in the "Python has stopped working" crash.

### Solution
This PR moves attribute cleanup to the proper lifecycle point - the Service destructor - with additional safety mechanisms:

**Key Changes:**

1. **Destructor-based cleanup**: `CleanupAttributes()` is now called from `Service::~Service()` instead of at the end of `Run()`. This ensures cleanup only happens when the object is truly being destroyed, not while it's still potentially accessible.

2. **Destruction state tracking**: Added `m_IsBeingDestroyed` atomic flag that is set at the start of the destructor, providing a clear signal that the Service is being torn down.

3. **Access guards during destruction**: `GetProperty()` and `SetProperty()` now check `m_IsBeingDestroyed` and throw descriptive errors if accessed during cleanup, preventing use-after-free scenarios.

4. **Ordered cleanup**: Properties are cleared first (releasing lambda captures of `this`), then Commands, then DataStreams. This ordering ensures that the most problematic circular references are broken first.

### Technical Details

**Original Implementation:**
```cpp
void Service::Run() {
    // ... service logic ...
    
    CleanupAttributes();  // Called while Service object still fully alive
}

Service::~Service() {
    // Empty - containers cleaned up by implicit member destructors
}
```

**Problems with Original:**
- Clearing containers while Service object is still active caused Windows access violations
- Race conditions if other threads/code still held references to properties/commands
- No way to detect if cleanup was in progress

**New Implementation:**
```cpp
void Service::Run() {
    // ... service logic ...
    
    // No explicit cleanup - let destructor handle it
}

Service::~Service() {
    m_IsBeingDestroyed = true;  // Signal destruction in progress
    CleanupAttributes();         // Break circular references safely
}

void Service::CleanupAttributes() {
    m_Properties.clear();   // Release lambda captures of 'this'
    m_Commands.clear();     // Release command references
    m_DataStreams.clear();  // Clean up remaining streams
}

PR #392, opened 2025-11-16
Move execution of commands and the service proxy obtaining the API to a message broker.

ehpor:
The dissemination of the API (names of properties, commands and data streams, the heartbeat stream id, configuration) and executing commands is currently done using the Server/Client C++ classes over TCP using ZeroMQ. This PR changes this to be done via the message broker instead. This is part of an ongoing transition from ZeroMQ to a fully message broker communication. After this PR, only the shutdown request to a service, and service registration at the testbed server is done via ZeroMQ.

### DIFF
--- a/catkit_core/Service.cpp
+++ b/catkit_core/Service.cpp
@@ -53,8 +53,7 @@ Service::Service(string service_type, string service_id, int service_port, int t
 
 	LOG_DEBUG("Registering request handlers.");
 
-	m_Server.RegisterRequestHandler("get_info", [this](const string &data) { return this->HandleGetInfo(data); });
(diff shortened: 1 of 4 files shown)


steigersg:
Pulled this branch on the HiCAT PC, reinstalled catkit2, started the server, and then was successfully able to open a service (TA camera), and started and stopped acquisition with no errors. @ehpor and @RemiSoummer, is this an adequate test of this PR? happy to run additional tests if needed

steigersg (approved):
Confirmed in person with @RemiSoummer that the test I performed is sufficient to test this PR- approved!

ehpor:
Confirmed with `mtop` that the expected messages are there.

PR #394, opened 2025-11-18
Bugfix/new 4d software compatibility 

### LINKED ISSUE #393: Need to add Accufiz Interferometer Entry Points to pyproject.toml
This includes the `accufiz_intereferometer_sim` service as well

steigersg:
Add compatibility for the new 4SightFocus software which is needed to use the WebServer on Windows 11. 

This involves the following changes:
- Update WebServer syntax when `new_software: True`
- Add CSV to FITS conversion since the new software cannot save HDF5 files
- add accufiz entry points to `pyproject.toml`

Open Question: Do we want to continue to use http requests? or should we think about migrating to the 4Sight python API? (likely out of scope for this particular PR)

To Do:
- [x]  Full test on hardware 

Closes #393.


lanemeier7:
I am working on a version that works directly with the BitFlow CameraLink and does the phase processing separately. 

I think one issue is that our camera and software is very old, and does not have the Python API availble. We do not have funding at the moment to upgrade it.

steigersg:
@lanemeier7 cool and thanks for keeping us in the loop! We would definitely be interested if that is something that is also compatible with our model/software though I still think I will proceed with this "patch fix" PR since we need it to take data with our 4D in the next couple of weeks. If we do go the way of the python API we will certainly not clobber any of the services that you need to run your interferometer (i.e. by making an accufiz_v2 or something)

lanemeier7 (commented):
This bug may have been introduced by a different update - but when I test this code on our 4D in our lab i need to modify the take_measurement function such that  is uses `np.ascontiguousarray`:
``` python
def take_measurement
...
return np.ascontiguousarray(image, dtype=np.float32)
```

lanemeier7:
Hi @steigersg , checking this branch out on our lab system, I'm perplexed because you have the `np.ascontinguous` in the return which fixed it for me before.

```
  File ".\exospec\tests\test_accufiz.py", line 17, in test_measurement
    img = af_proxy.take_measurement()
  File "C:\catkit2\catkit2\testbed\service_proxy.py", line 57, in cmd
    return self.execute_command(name, kwargs)
RuntimeError: Input array must be C continguous.
```

steigersg:
> Hi @steigersg , checking this branch out on our lab system, I'm perplexed because you have the `np.ascontinguous` in the return which fixed it for me before.
> 
> ```
>   File ".\exospec\tests\test_accufiz.py", line 17, in test_measurement
>     img = af_proxy.take_measurement()
>   File "C:\catkit2\catkit2\testbed\service_proxy.py", line 57, in cmd
>     return self.execute_command(name, kwargs)
> RuntimeError: Input array must be C continguous.
> ```

hmm that's strange, I know in your initial fix you had 

`return np.ascontiguousarray(image, dtype=np.float32)`

whereas I've implemented 

`return np.ascontiguousarray(image)`

so maybe including the dtype makes a difference?

lanemeier7:
ok - confirmed the dtype=np.float32 is important. When I add that in it works. But without it I get the "RuntimeError: Input array must be C continguous." message posted previously

steigersg:
> ok - confirmed the dtype=np.float32 is important. When I add that in it works. But without it I get the "RuntimeError: Input array must be C continguous." message posted previously

Interesting, I'll add that in then. I don't know enough about C contiguous arrays to know why that would be the case, but it doesn't impact anything on our end to include

steigersg:
actually a quick note @lanemeier7 - copying the image array to the contiguous array does have a time penalty that I know was a potential issue for us in the past with our science camera (which we want to run fast). I don't think we care about the time penalty as much with the 4D since the image saving/loading over the servers is pretty slow but something to keep in mind

lanemeier7 (approved):
this is working now in our lab. Thanks!

PR #396, opened 2025-11-27
Move service heartbeat and service state on broker

ehpor:
Depends on #392. Do not review before that PR is merged.

This PR moves the heartbeat and state of services onto the message broker. This is part of an ongoing effort to move as much information as possible on the message broker.


ehpor:
I ran into a bug where message headers were not cleaned up properly. This bug is, I think right now at least, unrelated to this PR. It however started when an infinite loop of publishing arrays (heartbeats) were done.

ivalaginja:
The code changes look globally good to me. @ehpor is there an appropriate hardware test this should be run in? And would this be better done after resolving the existing conflicts?

PR #397, opened 2025-12-04
Updating the VCP port (12/25). Adding test for catching behavior. 

RemiSoummer:
As discussed in #304  and associated PR #326 the MCLS1 can change port (COM and VCP) after a computer reboot. 
This is hicat-specific and will need to be fixed so we don't have to do those PRs anymore. 

This PR updates the current VCP port on HICAT and also adds another test to provide more robustness to catch this issue. 

In the past we were "lucky" that the port had moved to an unused one, so the test in place was able to catch the missing port and raise the error.   This time around the port switched to a port in used by another device, and this resulted in a ValueError when submitting to the data stream (the data was presented as empty string). 

The PR adds a test for a ValueError at the data stream submission to be able to catch the issue and make it immediately clear that this is the port jumping issue.


steigersg (approved):
confirmed LGTM

PR #398, opened 2025-12-04
Remove hardcoded vcp_port names and instead pull values from the config

### LINKED ISSUE #304: Migrate hardcoded VCP port in MCLS1 service

steigersg:
Gets a VCP port value from the config instead of assuming a given hardcoded value. Requires users not using the default of 'VCP0' to include a `vcp_port` in the config if using the MCLS1. Future port changes will then not require additional catkit2 PRs. 

Closes #304. 

For additional discussion see #303, #326 and #397

### DIFF
--- a/catkit2/services/thorlabs_mcls1/thorlabs_mcls1.py
+++ b/catkit2/services/thorlabs_mcls1/thorlabs_mcls1.py
@@ -105,6 +105,7 @@ class ThorlabsMcls1(Service):
 
         self.threads = {}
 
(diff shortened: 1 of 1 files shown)


RemiSoummer (approved):
Looks good

PR #401, opened 2025-12-10
Physik Stage Controller Service

### LINKED ISSUE #400: Physik Stage Controller Service
We are using a Physik Stage Controller at GSFC for tip tilt control and want to make it available to the general group

lanemeier7:
# Add Physik Instrumente Stage Controller Service

## Description

This PR adds a new service for controlling Physik Instrumente (PI) motion stages through the catkit2 framework. The service provides a thread-safe interface to PI hardware using the `pipython` library and follows the established catkit2 service patterns.

## Motivation

Physik Instrumente stages are commonly used in optical testbeds for precision motion control (tip-tilt mirrors, focus stages, etc.). This generic service allows any testbed to integrate PI hardware without duplicating hardware interface code.

### Data Streams

- `positions`: Current positions of all configured axes (float64 array)
- `target_positions`: Target positions when moves are commanded (float64 array)
- `is_moving`: tell if the stage is moving (bool)

### Properties

Dynamic properties created for each axis based on configuration:
- `position_{axis_name}`: Get/set position for each configured axis

### Commands

- `move_to(positions)`: Move to absolute positions
- `move_relative(deltas)`: Move relative to current positions
- `get_positions()`: Return current positions as dictionary
- `stop_motion()`: Emergency stop all motion

## Configuration Example

```yaml
physik_stage_controller:
  service_type: physik_stage_controller
  simulated_service_type: physik_stage_controller_sim
  requires_safety: false
  controller_name: 'E-727.3SDA'  # PI controller model
  stages: 'S-330.2SH'            # Stage model
  refmodes: 'FNL'                # Reference mode
  SN: '0123456789'               # Serial number (required)

# Map friendly axis names to hardware axis numbers
axis_map:
  x: 1
  y: 2

# Initial positions for each axis
initial_position:
  x: 0.0
  y: 0.0
```

```python
# Use from another service
class MyService(Service):
    def __init__(self):
        super().__init__('my_service')
        self.stage = getattr(self.testbed, 'physik_stage_controller')
    
    def do_something(self):
        # Move to absolute position
        self.stage.move_to({'x': 10.0, 'y': 20.0})
        
        # Move relative
        self.stage.move_relative({'x': 1.0})
        
        # Get positions
        pos = self.stage.get_positions()
```


catkit2/services/physik_stage_controller/physik_stage_controller.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+class PhysikStageController(Service):
+    """Service for controlling Physik Instrumente (PI) stages."""
+
+    def __init__(self):
+        super().__init__('physik_stage_controller')
+
+        self.pidevice = None
+        self.is_initialized = False
```
ehpor, line 14: I don't think these checks on `is_initialized` are strictly necessary, but they can't hurt so leave them in.

catkit2/services/physik_stage_controller/physik_stage_controller.py
```diff
@@ -0,0 +1,248 @@
...
+    def open(self):
+        """Initialize the PI device and connect."""
+        controller_name = self.config.get('controller_name', 'E-727.3SDA')
+        serialnum = self.config.get('SN', None)
+
+        if serialnum is None:
+            raise Exception('Must have controller serial number in config to find controller')
+
+        # Get initial positions from config
+        initial_pos = self.config.get('initial_position', {'x': 0, 'y': 0})
```
ehpor: This assumes x and y axis names. Is that intended?

catkit2/services/physik_stage_controller_sim/physik_stage_controller_sim.py
```diff
@@ -0,0 +1,229 @@
...
+        self.is_moving_stream.submit_data(np.array([0], dtype='int8'))
+
+        # Create properties for each axis
+        for name, axis_num in self.axis_map.items():
+            self._make_axis_property(name, axis_num)
+
+        # Create commands
+        self.make_command('move_to', self.move_to)
+        self.make_command('move_relative', self.move_relative)
+        self.make_command('get_positions', self.get_positions)
```
ehpor, line 75: Making these commands makes them relatively slow compared to data streams, up to around 1ms of latency at times. Will that be an issue?

docs/index.rst
```diff
@@ -42,6 +42,7 @@ Catkit2
    services/oceanoptics_spectrometer
    services/omega_ithx_w3
    services/phasics_cam
+   services/physic_stage_controller
```
ehpor: physik with a k.

ehpor (changes requested):
Just the typo is a required change. The other comments are more informational for you to be aware of.

docs/index.rst
lanemeier7: Figures the one change I made by hand is wrong ;-)

catkit2/services/physik_stage_controller/physik_stage_controller.py
lanemeier7: it is not a bad assumption, but could conflict in the event that axis names are defined and initial positions are not defined in configs

lanemeier7: One better thought: is if initial_positions is not specified - don't move anything
``` python
# Get initial positions from config
initial_pos = self.config.get('initial_position', None)

# if no intial positions specified, don't move anything
if initial_pos:
    # Initialize current positions
    for name, axis_num in self.axis_map.items():
        if name in initial_pos:
            self.current_positions[axis_num] = float(initial_pos[name])
        else:
            self.current_positions[axis_num] = 0.0

    self.log.info(f'Initial positions: {initial_pos}')
```

ehpor: Or maybe if the initial positions are not given, then take the keys from the `axis_map`?

ehpor: Sorry, I misread what your comment does. Yes, that seems nice. The inner if-else can be replaced by a `initial_position.get(...)`.

catkit2/services/physik_stage_controller_sim/physik_stage_controller_sim.py
lanemeier7: Thanks for bringing that to my attention, I'll check with my colleagues to see if this is an issue for us. We want to use this to control tip/tilt as fast as we can

lanemeier7: for now this is not an issue for us, what do you suggest if we need to move the stage faster?

ehpor: I'd use data streams directly. Have threads wait on new frames to come in on a data stream, then move to the requested position when that happens. You can look at the picomotors to see how that's done there.

But yes, even if it was an issue, I'd still merge this PR to get stuff started and people using that stage.

ehpor:
@lanemeier7 I think you forgot to push your commits.

catkit2/services/physik_stage_controller_sim/physik_stage_controller_sim.py
lanemeier7: I'm going to spend a little more time on this - would be nice to do it more or less in the best way possible.

ehpor (approved):

PR #404, opened 2026-01-28
Bugfix/service run cleanup

lanemeier7:
Ensures the GIL is held when CleanupAttributes() destroys Python callbacks.

Previously, the GIL was released for the entire run() call, but CleanupAttributes() should execute with the GIL held
```
.def("run", [](Service &service)
{                                      // GIL: RELEASED (by call_guard)
    service.Run(error_check_python);   // GIL: RELEASED
                                       //   ↳ internally calls CleanupAttributes()
                                       //   ↳ Py_DECREF() 💥 NO GIL!
}
, py::call_guard<py::gil_scoped_release>())
                                       // GIL: RE-ACQUIRED (too late)
```

Now we use explicit scoped release ensuring the GIL is re-acquired before cleanup.

``` C++
.def("run", [](Service &service)
{                                          // GIL: HELD
    {                                      // GIL: HELD
        py::gil_scoped_release release;    // GIL: RELEASED ←──┐
        service.Run(error_check_python);   // GIL: RELEASED    │ scope of
    }                                      // GIL: HELD ←──────┘ 'release'
    
    service.CleanupAttributes();           // GIL: HELD ✓ Safe!
})
```

This fixes issue #359

### DIFF
--- a/catkit2/bindings.cpp
+++ b/catkit2/bindings.cpp
@@ -562,8 +562,15 @@ PYBIND11_MODULE(catkit_bindings, m)
 		.def_property_readonly("config", &Service::GetConfig)
 		.def("run", [](Service &service)
 		{
-			service.Run(error_check_python);
-		}, py::call_guard<py::gil_scoped_release>())
+			{
+				py::gil_scoped_release release;
+				service.Run(error_check_python);
+			}
+			// CleanupAttributes must be called with GIL held because
+			// m_Properties and m_Commands contain Python callbacks wrapped
+			// in std::function. Destroying them without the GIL causes crashes.
+			service.CleanupAttributes();
+		})
 		.def("open", &Service::Open)
 		.def("main", &Service::Main)
 		.def("close", &Service::Close)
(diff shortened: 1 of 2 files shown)


lanemeier7:
@ehpor  does this explanation makes sense? I must admit I do have a little conceptual trouble with the GIL and C++ boundaries.

ehpor:
Yes. The `CleanupAttributes()` call can release Python resources, or at least modify their reference count. So in my mind, it totally makes sense that you need to hold the Global Interpreter Lock to avoid Python accessing those resources at the same time.

ehpor (approved):

lanemeier7:
Hi @steigersg if you can confirm that re-installing this branch version of catkit2, starting up your testbed then shutting the testbed down runs smoothly that would make me feel better.

steigersg:
> Hi @steigersg if you can confirm that re-installing this branch version of catkit2, starting up your testbed then shutting the testbed down runs smoothly that would make me feel better.

apologies for the delay on this @lanemeier7, was away fro conference travel last week but I will test this either this week or next week

steigersg (approved):
pulled and compiled this branch and then started our server, opened some services (camera, light source) and then shut everything down successfully so this looks good to me!

lanemeier7:
> pulled and compiled this branch and then started our server, opened some services (camera, light source) and then shut everything down successfully so this looks good to me!

sold!

Issue #402, opened 2025-12-12
Sub 1ms control with Physik Stage Controller


lanemeier7:
Use data streams instead of commands to increase the speed of the physik stage controller service

PR #407, opened 2026-02-12
Added a DataStream option for calling move_to. Issue #402

MichaelPhilbin:
Issue #402 : "Use data streams instead of commands to increase the speed of the physik stage controller service."

Currently, using make_command in catkit2 adds an overhead delay of 1–2 ms per command. To reduce this latency, we introduce a DataStream to handle move_to commands directly.

The new DataStream approach allows continuous submission of positions without the per-command overhead.

Changes:
 - Added a DataStream called move_to_stream, structured like the existing target_positions DataStream (float64 array).
 - move_to_stream has a worker thread that calls get_next_frame. (includes a timeout to avoid hanging)
 - Each frame (float64 array) is passed to the existing move_to logic.

Example call: 
```python
move = np.array([1.0, 2.0], dtype=np.float64)
physik_stage_controller.move_to_stream.submit_data(move)
```


MichaelPhilbin:
@lanemeier7 When you have a moment, could you review this and confirm that it behaves as expected?  
Also, I’d appreciate your feedback on whether the wait time seems appropriate.

lanemeier7:
Hi @MichaelPhilbin , this is great! I am hoping to test this out soon. In the mean time can you post the results of a benchmark where you clock the `physik_stage.move_to(x,y)` method vs the datastream method of commanding moves in the simulation? Note: you will likely need something like `time.perf_counter()` to measure at the granularity you need. Would be good to get a statistical average too, like the average move time of 100 or 1,000 moves to make it a robust benchmark.

lanemeier7:
> @lanemeier7 When you have a moment, could you review this and confirm that it behaves as expected? Also, I’d appreciate your feedback on whether the wait time seems appropriate.

The wait time is totally fine. It is not actually a 'wait' in the sense you are delaying anything, but how long to wait for the next datastream frame to come in. If the frame does not come out in 250ms ideally the code would catch the exception and try again quietly.

MichaelPhilbin:
> Hi @MichaelPhilbin , this is great! I am hoping to test this out soon. In the mean time can you post the results of a benchmark where you clock the `physik_stage.move_to(x,y)` method vs the datastream method of commanding moves in the simulation? Note: you will likely need something like `time.perf_counter()` to measure at the granularity you need. Would be good to get a statistical average too, like the average move time of 100 or 1,000 moves to make it a robust benchmark.

For sure @lanemeier7, here are some benchmarks:

I tested latency calling the moves for each method:
 - Command-based submit avg latency: 2.405 ms
 - Stream-based submit avg latency: 0.008 ms

I also tested the order of moves, to make sure there are no race conditions/improper order of moves:

Generated moves:
[[103.84524952 195.37225583]
 [389.01097397 335.97354249]
 [183.7983758  419.36134253]
 [178.85371529  60.56248491]
 [447.6063946   14.92429598]
 [131.09170347  32.29280508]
 [ 83.31291343 282.4238569 ]
 [365.77087681  82.6255176 ]
 [132.60818254  30.37564377]
 [226.18591608  85.66447073]
 [411.45943112 325.46012306]
 [128.33539807 201.86298074]
 [ 23.54666738 182.75924306]
 [150.73195092 269.60245339]
 [ 89.51862559  88.54234373]
 [220.58730652 355.53552746]
 [315.40291331 496.86938229]
 [326.84596972 486.13046317]
 [279.54426293 234.3876597 ]
 [287.27465396 387.64210862]]

--- Stream-based moves ---
Move 1: target=(103.85,195.37), current=(103.85,195.37)
Move 2: target=(389.01,335.97), current=(389.01,335.97)
Move 3: target=(183.80,419.36), current=(183.80,419.36)
Move 4: target=(178.85,60.56), current=(178.85,60.56)
Move 5: target=(447.61,14.92), current=(447.61,14.92)
Move 6: target=(131.09,32.29), current=(131.09,32.29)
Move 7: target=(83.31,282.42), current=(83.31,282.42)
Move 8: target=(365.77,82.63), current=(365.77,82.63)
Move 9: target=(132.61,30.38), current=(132.61,30.38)
Move 10: target=(226.19,85.66), current=(226.19,85.66)
Move 11: target=(411.46,325.46), current=(411.46,325.46)
Move 12: target=(128.34,201.86), current=(128.34,201.86)
Move 13: target=(23.55,182.76), current=(23.55,182.76)
Move 14: target=(150.73,269.60), current=(150.73,269.60)
Move 15: target=(89.52,88.54), current=(89.52,88.54)
Move 16: target=(220.59,355.54), current=(220.59,355.54)
Move 17: target=(315.40,496.87), current=(315.40,496.87)
Move 18: target=(326.85,486.13), current=(326.85,486.13)
Move 19: target=(279.54,234.39), current=(279.54,234.39)
Move 20: target=(287.27,387.64), current=(287.27,387.64)

I made and used this test file:
[test_physik.py](https://github.com/user-attachments/files/25275277/test_physik.py)

MichaelPhilbin:
I also changed the DataStream to do each requested move in order since when testing the order of moves it looked like it repeated the same move multiple times. @lanemeier7

This fixed the issue, but I am unsure if you want it to perform that way.

Edit: Never mind changed it back to NEWEST_ONLY and it works fine; I probably had an issue in my testing script originally.

lanemeier7:
Hey Michael,

This is a really great progress and catkit2 mastery on your end. I appreciate you taking on this difficult task so quickly!

What I am really after is the time it truly takes to theoretically send the MOV command inside the physik service either via function call which you clocked at 2.45ms or via the data stream triggering the physik service pidevice.MOV(axis_positions) command inside.. Which in my test case takes 15ms
``` python
nintervals = 100
# Now monitor position for nintervals additional intervals
print(f"\nMonitoring position for {nintervals} intervals...")

def move_and_wait(stage, newposx, newposy):
    # Define the target position
    target_move = np.array([newposx, newposy], dtype=np.float64)

    # Submit the move command
    stage.move_to_stream.submit_data(target_move)

    moving = True
    while moving:
        pos_frame = stage.positions.get_latest_frame().data
        if (pos_frame[0] == newposx) and (pos_frame[1] == newposy):
            moving = False

interval_start_time = time.perf_counter()
for i in range(100):
    try:

        move_and_wait(stage, i, i)

    except RuntimeError as e:
        print(f"Timeout at interval {i+1}: {e}")

final_monitoring_time = time.perf_counter()
total_monitoring_time = final_monitoring_time - interval_start_time

print(f"\n=== SUMMARY ===")
print(f"Time for 100 monitoring intervals: {1e3*total_monitoring_time:.3f} ms")
print(f"Average interval time: {1e3*total_monitoring_time/100:.4f} ms")
```

There are a couple potential flaws in the simulator you might need to research.. Is that it may be either slowed down by the simulated move logic that is in the sim code and not the non-sim version, or the main loop update rate.

Another thing to look out for is that the move_to_stream  might be redundant with target_positions stream. Perhaps we can use the target_positions datastream for this purpose instead since it already exists.

MichaelPhilbin:
> Hey Michael,
> 
> This is a really great progress and catkit2 mastery on your end. I appreciate you taking on this difficult task so quickly!
> 
> What I am really after is the time it truly takes to theoretically send the MOV command inside the physik service either via function call which you clocked at 2.45ms or via the data stream triggering the physik service pidevice.MOV(axis_positions) command inside.. Which in my test case takes 15ms
> 
> ```python
> nintervals = 100
> # Now monitor position for nintervals additional intervals
> print(f"\nMonitoring position for {nintervals} intervals...")
> 
> def move_and_wait(stage, newposx, newposy):
>     # Define the target position
>     target_move = np.array([newposx, newposy], dtype=np.float64)
> 
>     # Submit the move command
>     stage.move_to_stream.submit_data(target_move)
> 
>     moving = True
>     while moving:
>         pos_frame = stage.positions.get_latest_frame().data
>         if (pos_frame[0] == newposx) and (pos_frame[1] == newposy):
>             moving = False
> 
> interval_start_time = time.perf_counter()
> for i in range(100):
>     try:
> 
>         move_and_wait(stage, i, i)
> 
>     except RuntimeError as e:
>         print(f"Timeout at interval {i+1}: {e}")
> 
> final_monitoring_time = time.perf_counter()
> total_monitoring_time = final_monitoring_time - interval_start_time
> 
> print(f"\n=== SUMMARY ===")
> print(f"Time for 100 monitoring intervals: {1e3*total_monitoring_time:.3f} ms")
> print(f"Average interval time: {1e3*total_monitoring_time/100:.4f} ms")
> ```
> 
> There are a couple potential flaws in the simulator you might need to research.. Is that it may be either slowed down by the simulated move logic that is in the sim code and not the non-sim version, or the main loop update rate.
> 
> Another thing to look out for is that the move_to_stream might be redundant with target_positions stream. Perhaps we can use the target_positions datastream for this purpose instead since it already exists.

Hey @lanemeier7,

Thank you for the example test code and for clarifying what we are testing for.

Personally, I think keeping them separate is cleaner for a few reasons:

1. Avoiding circular logic:
If a user calls move_to() via the command (not the stream), we'd want to update the target_positions_stream telemetry. But if updating that telemetry also triggers the move command, we'd create an infinite loop. We could work around this with separate internal functions, but that adds complexity.

2. Clear separation of concerns:
Structurally, it makes sense to have:
 - Command stream (input): move_to_stream - how clients tell the service what to do
 - Telemetry stream (output): target_positions_stream - how the service reports its current state

That said, I don't have a lot of experience in software architecture, so if you have a different perspective or see flaws in this reasoning, I'm more than willing to learn from your experience.

As far as the simulated motion for the stage I am unsure of what values would be flawed. I did some research and a "E-727.3SDA controller with S-330.2SH stages" has a servo control of 20 kHz, at the moment the simulation is using 100 Hz. 

I also checked what the correct motion speed would be, and as far as I can tell piezo tip/tilt stages are not specified by a traditional motion speed (e.g., mm/s or deg/s) but rather by bandwidth and mechanical resonance. For the S-330.2SH, the mechanical resonance is on the order of ~1.6 kHz with load.

Because of the above, I replaced the velocity simulated movement with a bandwidth model (using a representative closed-loop bandwidth of ~400 Hz for the S-330.2SH). After these changes I still get about an average interval time of around 10-11 ms. 

Looking further into the code, it seems that even if I changed the refresh rate to 20 kHz, Windows has a limitation on its sleep granularity (~15 ms). The Service Sleep() method (in Util.cpp) uses std::this_thread::sleep_for(), which on Windows cannot achieve sub-millisecond precision. This results in an actual telemetry rate of about ~68 Hz instead of the configured 20 kHz.

I think realistically we can't simulate the real performance of the hardware using what we have. 
Please let me know if any of this is wrong.

What I referenced: [Document 1](https://usermanual.wiki/Pdf/E727xAPDatasheet.1243116320/view), [Document 2](https://www.pi-usa.us/en/products/fast-steering-mirrors-fsm-and-piezo-tiptilt-platforms-for-active-optics/s-330-fast-piezo-tiptilt-platform-300700/#specification)

Test File: [test_physik.py](https://github.com/user-attachments/files/25319576/test_physik.py)

lanemeier7 (dismissed):
Excellent work!

catkit2/services/physik_stage_controller/physik_stage_controller.py
```diff
@@ -81,54 +88,74 @@ def open(self):
...
+            with self.mutex:
+                return self.pidevice.qPOS()[str(axis_num)]
 
         def setter(value):
-            self.move_to({name: float(value)})
+            with self.mutex:
+                actual = self.pidevice.qPOS()
+            positions = {n: actual[str(num)] for n, num in self.axis_map.items()}
+            positions[name] = float(value)
+            self.move_to_wrapper(positions)
```
lanemeier7, line 113: I'd eliminate the position querying in the setter function. It adds an unnecessary query.

```
def setter(value):
    self.move_to_wrapper({name: float(value)})
```

lanemeier7 (changes requested):
Looks great! Only one minor change.

catkit2/services/physik_stage_controller/physik_stage_controller.py
MichaelPhilbin: Unfortunately, if we want the setter to go through the move_to_wrapper, it expects all axes due to the way we are passing moves through the DataStream. 

So, if we want the properties to update through the target DataStream we need to query for the other axis the property didn't send. If you want, we don't really need to make the properties go through the DataStream. If we didn't go through the DataStream the actual move_to function call can handle getting sent only one axis (what we had before).

Although, I think the delay will be minimal especially because I don't think anyone should be using properties for high latency moves.

When you get a chance, let me know what you think would be best.

lanemeier7: Great points. I rescind my change request. I have tested this out on our testbed and approve.

lanemeier7 (approved):

lanemeier7:
Yahoo! Thanks for making this change. We can run our control loops at 600 Hz now with our Physik thanks to this. Previously it was 100-150Hz max.

PR #412, opened 2026-02-17
Started the documentation for thorlabs_mcls1

### LINKED ISSUE #399: Missing service documentation
Some services don't appear to have documentation pages/weren't included in the big documentation push described in #190 (e.g thorlabs_mcls1)

Not sure if this is intentional and a won't fix, but wanted to bring it up just in case.

alexisyslau:
@raphaelpclt Would you mind help me with this please?

### DIFF
--- a/docs/services/thorlabs_mcls1.rst
+++ b/docs/services/thorlabs_mcls1.rst
@@ -1,12 +1,12 @@
 Thorlabs MCLS1
 ==============
-Thorlabs MCLS1 service contains software for controlling the `Thorlabs MCLS1`
-
-
-How?
+Thorlabs MCLS1 service contains software for controlling the `Thorlabs MCLS1` laser source.
+The communication is done through serial communication. 
+Thorlabs software suite is publicly available at https://www.thorlabs.com/software-pages/mcls1/
 
 .. note::
-    ????
+
+None.
 
 
 Configuration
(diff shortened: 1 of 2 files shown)


docs/services/thorlabs_mcls1.rst
```diff
@@ -0,0 +1,54 @@
+Thorlabs MCLS1
+==============
+Thorlabs MCLS1 service contains software for controlling the `Thorlabs MCLS1` laser source.
+The communication is done through serial communication. 
+Thorlabs software suite is publicly available at https://www.thorlabs.com/software-pages/mcls1/
+
+.. note::
+
+None.
+
```
ivalaginja: If there is no note, you can delete it.

docs/services/thorlabs_mcls1.rst
```diff
@@ -0,0 +1,54 @@
...
+        emission: 1
+        current_setpoint: 100
+        low_flux_current_setpoint: 35
+        channel: 2
+        target_temperature: 25
+
+        channels:
+            1: 640
+            2: 670
```
ivalaginja, line 29: Sincen we use numbers as channel names, maybe make one 640 and one with a random name of sorts.

ivalaginja (changes requested):

docs/services/thorlabs_mcls1.rst
raphaelpclt: Channel numbers are what we send to the source, like `set_channel(1)` (this is not a working code). 
The wavelength that comes after is for the proxy to yield the correct wavelength when asked.

docs/services/thorlabs_mcls1.rst
raphaelpclt: Done

docs/services/thorlabs_mcls1.rst
ivalaginja: Sure but the wavelengths are just names, right? We could have called them Max, Lucie and Peter instead of 640, 660 and 680 also, right?

RemiSoummer: Each MCLS1 can be populated with different lasers so the 640 and 670 are just examples of what we have on  HiCAT and others will have different setups

ivalaginja (approved):

PR #426, opened 2026-02-21
Allow out-of-order submission of messages in `LocalMessageBroker`.

ehpor:
> [!NOTE]
> This PR is still work in progress.

### Summary
This PR fixes out-of-order message handling in the local message broker by refactoring the message availability tracking from separate first/last atomics to a unified bitmap-based approach.

### The Problem
Previously, the message broker tracked message availability using two separate counters: `first_frame_id` (oldest available) and `last_frame_id` (newest available + 1). This worked for sequential publishing but couldn't handle out-of-order messages. For example, if message 5 arrived before message 4, there was no way to track that 5 was available while message 4 was still missing.

### The Solution: Bitmap-Based Availability Tracking

Instead of tracking just a range `[first, last)`, we now track **which specific messages** in the circular buffer are available using a bitmap packed into a single atomic variable.

**How it works:**

The `availability` field is a 64-bit atomic containing two pieces of information:
- **Upper 48 bits**: The `last` message ID (the next ID that will be assigned). These 48 bits still allow us to publish messages for ~9 years at 1 million messages/sec.
- **Lower 16 bits**: A bitmap indicating which of the previous 16 messages are actually available.

**Bitmap interpretation:**
- Bit 0 = message `last-1` (the newest message)
- Bit 1 = message `last-2`
- Bit 2 = message `last-3`
- ...and so on

**Example:**
If `last = 10` and the bitmap is `0b0000000000010101` (binary):
- Bit 0 set → message 9 is available
- Bit 2 set → message 7 is available  
- Bit 4 set → message 5 is available
- Messages 6 and 8 are **not** yet available (out of order!)

This allows the broker to correctly handle out-of-order message publishing while maintaining lock-free concurrent access between publishers and subscribers.

Note: packing everything in a single 64bit atomic allows us to atomically update both at the same time. We can perform additional operations using CAS loops (done in `TryReserve()`, `TryReserveNext()` and `TryMakeAvailable()`.

### Key Changes

1. **Refactored TopicHeader availability tracking** Converted from separate `first_frame_id`/`last_frame_id` atomics to a single `availability` atomic containing packed `last_id` and 16-bit bitmap.

2. **Added C++17 bit operations** Implemented `countl_zero` and `countr_zero` functions (backported from C++20) in `Util.h/Util.inl` to efficiently count leading/trailing zeros in the bitmap.

3. **Moved message ID to MessageHeader** This is required since each nested topic has a different message ID and we need to track all.

4. **API improvements:**
   - Changed `GetOldestMessageId()`/`GetNewestMessageId()` to `GetBegin()`/`GetEnd()` (half-open range semantics: `[begin, end)`)
   - Added `TryReserve()`, `TryReserveNext()`, `TryMakeAvailable()` methods for atomic reservation and publishing
   - Requiring subscriptions to access messages (removed duplicate access path that bypassed subscription tracking)


ehpor:
One bug remaining: no messages are published on the root ("") topic, even though they should.

tests/test_message_broker.py
```diff
@@ -42,34 +35,34 @@ def test_message_subscription(broker):
...
     # Creating a NewestOnly subscription starting at ID 1 should return the newest message (ID 2).
-    subscription_newest2 = broker.subscribe(topic, preferred_next_frame_id=1, mode=MessageSubscriptionMode.NewestOnly)
+    subscription_newest2 = broker.subscribe(topic, preferred_next_message_id=1, mode=MessageSubscriptionMode.NewestOnly)
     m = subscription_newest2.get_next_message(0.01)
-    assert m.frame_id == 2
+    assert m.message_id == 2
 
     # A Sequential subscription starting at ID 1 should return the second message (ID 1).
-    subscription_sequential2 = broker.subscribe(topic, preferred_next_frame_id=1, mode=MessageSubscriptionMode.Sequential)
+    subscription_sequential2 = broker.subscribe(topic, preferred_next_message_id=1, mode=MessageSubscriptionMode.Sequential)
```
alexisyslau: Adding a test for out-of-order message handling? The current test only publishes messages sequentially (I think?), so it would be good to have this test to cover what this PR is trying to fix.

catkit_core/MessageBroker.cpp
```diff
@@ -4,8 +4,13 @@
 
 #include "LocalMessageBroker.h"
 
-Message::Message(MessageHeader *header, void *payload, std::uint64_t frame_id, bool has_been_published)
-	: m_Header(header), m_Payload(payload), m_FrameId(frame_id), m_HasBeenPublished(has_been_published)
+size_t GetTopicDepth(std::string_view topic)
+{
+	return std::count(topic.begin(), topic.end(), '/');
```
alexisyslau: Sooooo here `""` and `"topicA"` have the same depth, right? The count only increases when we have `/` like `"topicA/topicB"`. Maybe that’s why messages aren’t being published to the root topic properly?

alexisyslau: Then we need a +1 somewhere when counting the depth? 
"" : 0, "TopicsA": 1, "TopicsA/TopicB": 2

catkit_core/LocalMessageBroker.cpp
```diff
@@ -344,118 +593,119 @@ Message LocalMessageBroker::PrepareMessageImpl(std::string_view topic, size_t pa
...
+					// we will never be successful. This is only an error if this
+					// is the deepest topic.
+					if (i == 0)
+						throw std::runtime_error("Cannot reserve slot in topic.");
 
-		m_MessageHeaderAllocator->Acquire(message_header_index);
-		allocator->Acquire(message.m_Header->payload_info.block_handle);
+					break;
+				}
+			}
```
alexisyslau: It's a bit too complex for me to follow, just thinking out loud here:
1. `reserve_result.success` is true  --> message publish
2. reserve_result fails --> if can try again & not root --> skip the publishing and go to the next topic
3. reserve_result fails --> if can try again & root --> throw error
4. reserve_result fails --> can try again fails --> continue to publish the message? If yes, is this intended?

catkit_core/LocalMessageBroker.h
alexisyslau: TODO for this PR?

ehpor: Nope.

PR #437, opened 2026-05-04
Testbed Reload Configurations Mechanism

### LINKED ISSUE #403: Reloading Service Configurations
It would be nice to be able to reload testbed service configurations while the testbed is running. This would reduce the need to restart the testbed process for new configurations.

lanemeier7:
Adds a `reload_config` mechanism to the testbed so configuration can be updated without restarting the entire testbed process.

This does NOT force running services to use the new configuration. However when a service starts for the first time after configuration is reloaded, or individual service is restarted it will obtain the new configuration from the testbed.


proto/service.proto
ehpor, line 1: This file might've been committed accidentally?

catkit2/testbed/testbed.py
```diff
@@ -686,6 +697,67 @@ def register_service_type(self, service_type, path):
...
+
+        # Update the config
+        old_config = self.config
+        self.config = new_config
+
+        # Update simulation mode if it changed
+        if 'testbed' in new_config and 'simulated' in new_config['testbed']:
+            self.is_simulated = new_config['testbed']['simulated']
+
+        # Check for added/removed services
```
ehpor: You're just checking new/removed services, but not updating the dependency graph. A correct dependency graph is required for correct shutdown of the testbed.

catkit_core/TestbedProxy.cpp
```diff
@@ -197,6 +197,26 @@ void TestbedProxy::ShutDown()
...
+	{
+		reply.ParseFromString(MakeRequest("reload_config", Serialize(request)));
+	}
+	catch (...)
+	{
+		throw std::runtime_error("Unable to reload config.");
+	}
+
+	// Invalidate cached config so it gets refetched on next access
+	m_HasGottenInfo = false;
```
ehpor, line 221: This variable is _just for this instance_ of a `TestbedProxy`. It won't affect other `TestbedProxy` objects. I think you're aware of this, but I wanted to make sure.

catkit2/testbed/testbed.py
```diff
@@ -686,6 +697,67 @@ def register_service_type(self, service_type, path):
...
+                service_info = new_config['services'][service_id]
+                service_type = service_info['service_type']
+
+                if self.is_simulated and 'simulated_service_type' in service_info:
+                    service_type = service_info['simulated_service_type']
+
+                dependencies = service_info.get('depends_on', [])
+                self.services[service_id] = ServiceReference(service_id, service_type, ServiceState.CLOSED, dependencies, self.message_broker)
+
+        if removed:
```
ehpor: Maybe we could also have a warning for services that are running of which the config has changed?

ehpor (changes requested):

proto/service.proto
lanemeier7: oops! thanks for catching that

catkit2/testbed/testbed.py
lanemeier7: I have been using this mechanism to update flat maps, tip tilt stage position defaults, camera offset_x, offset_y defaults in case we want to start / stop the services and have them  come up the the new default without restarting the testbed. Since this is all tracked in the services.yml config file I can track it on git. Generally speaking when I hit reload I want to make sure the new configs are in place. A warning may over alert the user. Maybe info level would be a good way to status what has changed. I could go either way though

ehpor: Agreed on `INFO`. I meant "warning" as in "some kind of log message" rather than "a warning log message". I could've been more clear on that.

proto/service.proto
ehpor: This file is still there. It wasn't accidental?

lanemeier7: apologies - i meant to remove it - it somehow came back. I've put these on a different branch for a different PR - since I need some of these service mods for the GUI I provided to catkit2-hci

catkit2/testbed/testbed.py
```diff
@@ -686,6 +665,251 @@ def register_service_type(self, service_type, path):
         '''
         self.service_type_paths[service_type] = path
 
+    def _resolve_service_type(self, service_info):
```
ehpor, line 668: This should be a static_method, right?

catkit2/testbed/testbed.py
```diff
@@ -686,6 +665,251 @@ def register_service_type(self, service_type, path):
...
+            service defines a ``simulated_service_type``, that type is used instead.
+        '''
+        service_type = service_info['service_type']
+
+        if self.is_simulated and 'simulated_service_type' in service_info:
+            service_type = service_info['simulated_service_type']
+
+        return service_type
+
+    def _get_safety_service_id(self, config):
```
ehpor: This should be a static_method, right?

catkit2/testbed/testbed.py
```diff
@@ -686,6 +665,251 @@ def register_service_type(self, service_type, path):
...
+            The id of the safety service, or None if no safety service is configured.
+        '''
+        testbed_section = config.get('testbed', {})
+
+        if 'safety' in testbed_section:
+            return testbed_section['safety']['service_id']
+
+        return None
+
+    def _compute_reverse_dependencies(self, nodes, safety_service_id):
```
ehpor: This should be a static_method, right?

tests/test_testbed_dependency_graph.py
```diff
@@ -0,0 +1,105 @@
...
+from catkit2.testbed.testbed import Testbed
+
+
+class DummyTestbed:
+    '''A stand-in for a Testbed instance to call _compute_reverse_dependencies() on.'''
+    log = logging.getLogger(__name__)
+
+
+def compute(nodes, safety_service_id=None):
+    return Testbed._compute_reverse_dependencies(DummyTestbed(), nodes, safety_service_id)
```
ehpor: If this method was static, you wouldn't need to create a dummy testbed for testing.

catkit2/testbed/testbed.py
```diff
@@ -715,6 +939,24 @@ def start_service(self, service_id):
...
+            requires_safety = service_info.get('requires_safety', False)
+
+            if (service_type, dependencies, requires_safety) != (reference.service_type, reference.dependencies, reference.requires_safety):
+                reference.service_type = service_type
+                reference.dependencies = dependencies
+                reference.requires_safety = requires_safety
+
+                # Rebuild the dependency graph so shutdown ordering stays correct.
+                self._build_dependency_graph()
+
```
ehpor, line 961: I'm really confused what this is supposed to be doing. No need to add code comments, just need to explain in text here.

So, for running services, their config is "deferred" meaning that the (private) configuration of that service reference in the testbed server is not changed. And then only when you restart a service, it checks whether the config in the testbed config is up-to-date with the config stored in the service reference, and if it isn't, it updates it, and starts the service after.

What I don't understand is why there needs to be a separate path for running services and non-running services. Can't all services share this code path, no matter if the service is running or not? Also, why does not updating the private service reference inside the testbed server, which no one else can read, matter? Is it solely for the shutdown procedure?

I also have my concerns about partial rebuilding of the dependency graphs, where if you have two services that is updated, the dependency graph is required to be valid when you only update one of the two (ie. the one which you happen to restart first). This is a tiny edge case however, and I don't think it's worth any time pondering on now. Leave that for the poor soul who encounters that in the future, if there will be any.

ehpor (changes requested):

ehpor:
Also @lanemeier7 since you're clearly using AI to generate this PR, what model are you using? This needs to be documented for the JOSS paper.

lanemeier7:
> Also @lanemeier7 since you're clearly using AI to generate this PR, what model are you using? This needs to be documented for the JOSS paper.

I am using Claude Opus

catkit2/testbed/testbed.py
lanemeier7: Apologies on late responses. Doing a lot of multi tasking these days. Since the cat is out of the bag with Claude - here is the response from Claude:

> The idea is that `self.config` is *desired* state while a `ServiceReference` is a description of the *process that is actually running*. That's why a running service's reference isn't touched: its process was spawned under the old config, so `service_type` is the type of the live process and `dependencies` are the links that process actually holds. Overwriting those while the process is alive would make the reference describe a service that doesn't exist.
> 
> It's not only shutdown, though shutdown is the biggest consumer. The things that read a reference's copy:
> 
> - `dependencies`/`requires_safety` → `depended_on_by` → shutdown ordering (`testbed.py:1121`, `:1143`)
> - `dependencies` → eager start of dependencies (`testbed.py:1053`)
> - `service_type` → the path we spawn (`testbed.py:963`), the type we advertise to clients in `on_get_service_info` (`testbed.py:611`), and the equality check against what the process reports in `on_register_service` (`testbed.py:624`)
> 
> So it isn't purely private — `on_get_service_info` would hand clients a type that doesn't match the running process, and the registration check would reject a service whose reference had drifted from what was spawned.
> 
> On the separate paths: you're right, they should be one, and the shared one has to be the deferred one — eager application is the direction that's impossible for running services, while deferred application is already correct for stopped ones (a stopped service is exactly a service that will be spawned next, and `start_service` syncs the reference immediately before spawning). The `else` branch in `reload_config` (`:890-894`) is redundant with the block in `start_service`: every stopped service it eagerly updates would be updated again before it's spawned. I'll drop it, so the rule becomes: references are synced from config at spawn time, and `depended_on_by` is always computed from the references. `reload_config` keeps the prospective-graph computation for *validation only* (so an invalid config is still rejected atomically with the testbed untouched) and rebuilds the installed graph from the references.
> 
> One consequence worth stating: with the eager branch gone, the graph installed at reload keeps the stopped service's old edges until it starts. That's fine for shutdown ordering — a stopped service has none — and it removes the current inconsistency where a reference's `dependencies` and its `depended_on_by` come from different configs.
> 
> Agreed on the partial-rebuild edge case; leaving it.

catkit2/testbed/testbed.py
lanemeier7: Done in 7b9eda1 — now a `@staticmethod` taking `is_simulated` explicitly, with a small test added.

catkit2/testbed/testbed.py
lanemeier7: Done in b656516.

catkit2/testbed/testbed.py
lanemeier7: Done in b656516.

tests/test_testbed_dependency_graph.py
lanemeier7: Done in b656516 — the dummy testbed is gone, the test calls the static method directly.

catkit2/testbed/testbed.py
lanemeier7: Unified in 7b9eda1. The `else` branch in `reload_config()` is gone, so references are only synced from the config in `start_service()` right before spawning, whether the service was running at reload time or not. `reload_config()` now validates the prospective graph, applies the config, adds/removes references, and rebuilds the installed graph from the references via `_build_dependency_graph()`. The partial-rebuild edge case is left as discussed.

catkit_core/TestbedProxy.cpp
lanemeier7: Yes, aware — this only invalidates this proxy's cache; other `TestbedProxy` instances refetch on their next access as usual.

ehpor (approved):
This looks good to me now. But I'm sure there's edge cases that we'll find when we start using this feature.

PR #438, opened 2026-05-05
Doc clarifications

ivalaginja:
This PR contains a couple of doc updates and simulator example updates I found worthy changing while I was working on capsule setup over the first couple of days.

- Updates the Getting started instructions to be more explicit about how to handle the cookiecutter results
- Adds info to the ZWO camera documentation about the required environment variable that points to the vendor SDK
- Inserts an explicitly stated unit wavelength with its own variable in the template simulator provided by cookiecutter, as it is otherwise just silently assumed and hard to identify for necessary changes.
- Both for cookiecutter template simulator, as well as simple simulator: changes the `camera_readout()` function to actually use the passed in camera_name variable and not the hard-coded `science_camera`. Also removes the `images` data stream from the simulator as image data streams are standardly associated with a (simulated or hardware) camera.

### DIFF
--- a/catkit2/services/simple_simulator/simple_simulator.py
+++ b/catkit2/services/simple_simulator/simple_simulator.py
@@ -58,7 +58,6 @@ class SimpleSimulator(Simulator):
 
             camera.images.submit_data(image)
         except Exception as e:
-        except Exception as e:
             self.log.error(str(e))
 
 if __name__ == '__main__':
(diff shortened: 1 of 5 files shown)


steigersg (approved):
Looks good to me! However @RemiSoummer also wanted to look it over before merging

ivalaginja:
Pinging @RemiSoummer to see if we can merge this

RemiSoummer (approved):
looks great

PR #440, opened 2026-05-07
Clean up error handling in Newport XPS service

ivalaginja:
The error handling was a bit ambiguous so I separated out an error if there is no correct env variable set, or if it can simply not import the required module.

Tested on hardware and I was able to trigger both error messages independently.


ivalaginja:
Just a ping to reviewers that this PR is ready for review.

RemiSoummer (approved):
looks great

PR #441, opened 2026-05-08
thorlabs_mcls1: add OS-based backend selection (Windows UART lib / Linux serial)

ivalaginja:
This PR updates the `ThorlabsMcls1` service to support both Windows and Linux/macOS communication paths while keeping a single service implementation.

The MCLS1 service previously depended on the Thorlabs UART library flow, which is Windows-specific. We need the same service to work on non-Windows systems using direct serial communication (e.g., Linux).

I added runtime backend selection based on OS, and refactored all necessary methods to call one or the other, depending on the backend.

Backward-compatible for existing Windows setups (same UART library behavior).
Adds non-Windows support without introducing a second service class.

### TODO
- [x] Test on Linux machine (capsule)
- [x] Test on Windows machine


lanemeier7:
@ivalaginja I also recently used an mcls1 and made some changes to work directly with pyserial nearly the same time you started working this. I can test your changes on Windows

RemiSoummer (approved):
Looks good and tested on both linux and windows

PR #446, opened 2026-07-16
ALPAO DM service 

raphaelpclt:
This is just the code used at MPIA to control the ALPAO DM, and is here with the intention to start discussion among all the catkit2 / ALPAO users who might have something more involved. 

The current implementation is trying to mimic the BMC services, but merged the base service with the hardware implementation. This will have to be addressed. 

ALPAO's sdk takes as an input a list of normalized voltages ranging from -1 to 1, 1 being the maximum deviation it can produce. I have attempted to implement a version where, without gain map calibration provided, the service will interpret commands as normalized voltage. With a gain map, commands would be provided in the unit of the gain map. 

Finally this code uses the ALPAO SDK, that is known to introduce higher latency and latency jitter than what the DM can do. I have heard of an alternative control method bypassing the SDK and addressing the DM at a lower level for a faster and more deterministic control - this is out of the scope of this PR but might be something coming later.

### DIFF
--- a/catkit2/services/alpao_deformable_mirror/alpao_deformable_mirror.py
+++ b/catkit2/services/alpao_deformable_mirror/alpao_deformable_mirror.py
@@ -33,7 +33,7 @@ class AlpaoDeformableMirror(DeformableMirrorService):
         self.device = DM(self.serial_name)
         if not self.device.Check():
             raise RuntimeError(f'Could not connect to DM {self.serial_name}')
-        
+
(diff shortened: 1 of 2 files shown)


raphaelpclt:
Fyi @kyohoon-ahn, I have tried to merge our 2 implementations, but staying as close as possible to the BMC version. I still need to polish things and test before making it ready for review, but it's getting close to an actual version.

Issue #447, opened 2026-07-16
Importlib metadata entry point syntax change with python 3.10


raphaelpclt:
For the users on Python versions 3.10 and later, there seems to be a syntax change in the way to retrieval of the entry point, as described here https://docs.python.org/3.10/library/importlib.metadata.html

Here is a snapshot of the error message. 
<img width="925" height="98" alt="Image" src="https://github.com/user-attachments/assets/9522eeee-8f3b-4b96-ad80-c9b39bcc1f62" />

Changing `importlib_metadata.entry_points().get("catkit2.services", [])` to `importlib_metadata.entry_points(group="catkit2.services")` or equivalently `importlib_metadata.entry_points().select(group="catkit2.services")` solves the issue.

PR #448, opened 2026-07-16
Fix entry points retrieval syntax in newer python versions

raphaelpclt:
This addresses #447 

I am not sure of the retro compatibility with python 3.7, and I do not have an installed 3.7 to test, so I'll leave it as a draft pull request. EDIT: not compatible with older versions, as the CI tests kindly pointed out.

Also this might now fail if the list of entry point is empty since it was previously returning an empty list if no entry point was found.


ehpor:
Ahhh, sorry! I didn't see this PR and created an almost identical one: #462.

raphaelpclt:
Taken over by #462

PR #450, opened 2026-07-20
Add camera base class to service doc index

ivalaginja:
Because forgotten in #131 😬


steigersg (approved):
LGTM! Only comment is that we may want to consider adding an extra level to the index to collect all the base class services in the docs.

PR #451, opened 2026-07-22
Remove hard-coded sci cam flip in HiCAT ZWO camera (Revert #132)

ivalaginja:
Reverting #132 after we switched HiCAT over to use the v2 service for the ZWO cameras, which inherit from the new camera base class (#131) that includes frame orientation handling.


RemiSoummer (approved):

RemiSoummer:
tested on hicat

